PR #3, opened 2018-03-23
Rename packages

jhernand:
This patch adapts the project to use the name that corresponds to the
new git repository: openshift/autoheal.


jhernand:
@elad661, @zgalor, @moolitayer, @ironcladlou please review.

manifests/healer-deployment.yml
```diff
@@ -31,6 +31,6 @@ spec:
       serviceAccountName: healer
       containers:
       - name: healer
-        image: openshift-monitoring/alert-healer:0.0.0
+        image: autoheal/alert-healer:0.0.0
```
ironcladlou: may not matter so much here since the actual manifest used to install will live in the ansible role, but the name will probably need to follow our latest conventions, e.g. `openshift/origin-autoheal`

jhernand: Yes, I just wanted to keep this consistent with this particular change, as the `build` script that we use currently to build generates that image name. In the next pull request I will adapt everything to the standard Go tooling, including this file.

PR #5, opened 2018-03-25
Compare Go versions numerically

elad661:
This is a backport of https://github.com/openshift/release/pull/713

Original commit message:

Currently the `setup_env` function compares the required and available
Go versions alphabetically.  This doesn't work correctly when comparing
1.9 and 1.10. As a result it isn't possible to build with Go 1.10. To
avoid that issue this patch changes the function so that it compares the
versions numerically, using the `--version-sort` option of the `sort`
command.


elad661:
cc @jhernand

jhernand:
Looks good, thanks @elad661.

PR #6, opened 2018-03-26
Read configuration from yaml file

jhernand:
This patch changes the auto-heal service so that it reads the
configuration from a `autoheal.yml` configuration from a YAML file,
including the set of rules that used to be obtained with a Kubernetes
API query.

An example configuration file looks like this:

```yaml
#
# This section contains the details needed to connect to the AWX or Ansible
# Tower server that will execute the Ansible playbooks.
#
awx:
  #
  # The URL of the AWX API endpoint, including the `/api` suffix, but not the
  # `/v1` or `/v2` suffixes.
  #

  #
  # The URL of the proxy servere to use to connect to the AWX API. If not
  # present or empty then no proxy server will be used.
  #

  #
  # Reference to the Kubernetes secret that contains the user name and password
  # to use to connect to the AWX API.
  #
  credentialsRef:
    namespace: my-namespace
    name: my-awx-credentials

  #
  # Reference to the Kubernetes secret that contains the TLS certificates and
  # keys to use to connect to the AWX API. Usually this will connect only the
  # `ca.crt` entry, which should contain the PEM encoded certificates of the
  # authorities that will be trusted when verifying the certificates presented
  # by the AWX server. If not present or empty then the global system trusted
  # certificates will be used.
  #
  tlsRef:
    namespace: my-namespace
    name: my-awx-tls

  #
  # The name of the AWX project that contains the auto-heal job templates.
  #
  project: "My project"

#
# This section contains the healing rules.
#
rules:

#
# This rule runs an the `Start Node` AWX job when the `NodeDown` alert
# is fired. The playbook that handles this alert is expected to take the
# name of the down from the `node` variable, so we need to translate the
# `instance` label into the `node` variable passed in the `extraVars`.
#
- metadata:
    name: start-node
  labels:
    alertname: "NodeDown"
  awxJob:
    template: "Start node"
    extraVars: |-
      {
        "node": "{{ $labels.instance }}"
      }

#
# This rule runs a batch job that uses Python to say hello when the
# `NewFriend` alert is fired. The alert is expected to have a `name`
# label containing the name of the new friend.
#
- metadata:
    name: say-hello
  labels:
    alertname: "NewFriend"
  batchJob:
    apiVersion: batch/v1
    kind: Job
    metadata:
      namespace: default
      name: hello
    spec:
      template:
        spec:
          containers:
          - name: python
            image: python
            command:
            - python
            - -c
            - print("Hello {{ $labels.name }}!")
          restartPolicy: Never
```


jhernand:
@zgalor, @moolitayer, @elad661, @ironcladlou please review.

autoheal.yml
```diff
@@ -0,0 +1,108 @@
...
+#
+awx:
+  #
+  # The URL of the AWX API endpoint, including the `/api` suffix, but not the
+  # `/v1` or `/v2` suffixes.
+  #
+  address: https://my-awx.example.com/api
+
+  #
+  # The URL of the proxy servere to use to connect to the AWX API. If not
```
moolitayer: typo: server

moolitayer: `The URL of the proxy server used to`

README.md
```diff
@@ -1,8 +1,200 @@
...
-playbooks to resolve the root cause.
+This project contains the _auto-heal_ service. It receives alert notifications
+from the Prometheus alert manager and executes Ansible playbooks to resolve the
+root cause.
+
+## Configuration
+
+Most of the configuration of the auto-heal service is kept in a YAML
+configuration file.  The name of the configuration file is specified using the
+`-config-file` command line option. If this option isn't explicitly given then
```
zgalor: I am not 100% sure but I think that there is an openshift standard on using `--` as a long flag prefix: `--config-file`
@ironcladlou ?

README.md
```diff
@@ -1,8 +1,200 @@
...
+    VkVzdzI3Sm5BYlMyejNhbUF0YTc1QmNnVGcvOUFCdDV0VVc2VTJOKzkKbXc9PQotLS0tLUVORCBD
+    ...
+  tls.key: ""
+  tls.crt: ""
+```
+
+The `ca.crt` key should contain the certificates of the authorities that are for
+verifying the certificate presented by the AWX server.
+
+The `tls.key` and `tls.crt` are only used when it is necessary to use a client
```
elad661: Do we really want to support using client certificate for AWX auth?

README.md
```diff
@@ -1,8 +1,200 @@
...
+The `-kubeconfig` command line option is used to specify the location of the
+Kubernetes client configuration file. When running outside of a Kubernetes
+cluster the auto-heal service will use `$HOME/.kube/config` by default, the same
+used by the `kubectl` command. When running inside a Kubernetes cluster it will
+use the configuration that Kubernetes mounts automatically in the pod file
+system. So in most cases this command line option won't have to be explicitly
+included.
+
+The `-alsologtostderr` option is very convenient when running the auto-heal
+service, both in development and production environments.
```
moolitayer: Might be helpful to always log to stdout (so we could get the logs with `oc logs`)

README.md
jhernand: Probably not.

README.md
jhernand: I agree. We can do that in a follow-up patch.

README.md
```diff
@@ -1,8 +1,200 @@
...
+service will internally decide which version to use.
+
+The `proxy` parameter is optional, and it indicates what HTTP proxy should be
+used to connect to the AWX API. If this parameter is not specified, or if it is
+empty, then the connection will be direct to the AWX server, without a proxy.
+
+The `credentialsRef` parameter is a reference to the [Kubernetes
+secret](https://kubernetes.io/docs/concepts/configuration/secret) that contains
+the user name and password used to connect to the AWX API. That secret should
+contain the `username` and `password` keys. For example:
```
moolitayer: consider replacing the above section with a link to autoheal.yml so we have to maintain only one file.

pkg/config/loader.go
```diff
@@ -0,0 +1,248 @@
...
+		for _, file := range files {
+			l.files = append(l.files, file)
+		}
+	}
+	return l
+}
+
+// Load loads the configuration files and returns the resulting configuration object.
+//
+func (l *Loader) Load() (config *Config, err error) {
```
nimrodshn: @jhernand I suggest taking a look at viper (https://github.com/spf13/viper), it has all the config file logic you have implemented covered.

README.md
elad661: In this case we should probably not mention it in the docs, and only add it if someone asks for it specifically :)

ironcladlou:
The app needs to to be able to load serialized configuration from a ConfigMap rather than a file on disk. I personally would prefer a monolithic secret to tree of ConfigMap->(Secrets) with references for simplicity's sake right now (just like AlertManager configuration), but I'm not going to argue strongly for it.

README.md
ironcladlou: Should be using spf13/cobra like everything else

ironcladlou:
Oh sorry, was the intent to read from disk and mount the configmap with a volume?

ironcladlou:
I need more ☕️

pkg/config/loader.go
jhernand: Thanks @nimrodshn, that is very interesting, I will consider it for later. For now I think that this configuration mechanism is enough.

jhernand:
Yes @ironcladlou, the intent is to mount the config map and then read from the file.

pkg/config/loader.go
nimrodshn: I'll make a PR  ;)

README.md
jhernand: I agree with using `spf13/cobra`, but this pull request is already very large. We will address this in a separate pull request. I created #7 to track that.

README.md
jhernand: @moolitayer I opened issue #9 to track that.

README.md
jhernand: I agree, will do.

autoheal.yml
jhernand: Done.

jhernand:
The new version of the pull request addresses some of the comments. The rest are now tracked as issues.

elad661:
As future work is going to depend on this PR (for example, my work on TLS support), and since nobody had any clear objections, I think we should merge this now

moolitayer:
I planed to review the code today and use the opportunity to get to know it better.
I think we can afford to wait another day.

I'm taking longer then usually as I have also changed my idea recently, sorry.

cmd/autoheal/alerts_worker.go
```diff
@@ -88,9 +88,17 @@ func (h *Healer) startHealing(alert *alertmanager.Alert) error {
 	activated := make([]*monitoring.HealingRule, 0)
 	h.rulesCache.Range(func(_, value interface{}) bool {
 		rule := value.(*monitoring.HealingRule)
-		if h.checkConditions(rule, alert) {
+		ok, err := h.checkRule(rule, alert)
```
moolitayer: ok usually means something can [succeed or fail]](https://tour.golang.org/methods/15)
 Maybe `matches`?

cmd/autoheal/alerts_worker.go
```diff
@@ -88,9 +88,17 @@ func (h *Healer) startHealing(alert *alertmanager.Alert) error {
...
+		if err != nil {
+			glog.Errorf(
+				"Error while checking if healing rule '%s' matches action '%s': %s",
```
moolitayer: Should this say `matches alert` instead of `matches action`?

cmd/autoheal/alerts_worker.go
jhernand: Yes, `matches` makes sense, will change it.

cmd/autoheal/alerts_worker.go
jhernand: Yes, I will change it.

autoheal.yml
```diff
@@ -0,0 +1,105 @@
...
+
+#
+# This section contains the healing rules.
+#
+rules:
+
+#
+# This rule runs an the `Start Node` AWX job when the `NodeDown` alert
+# is fired. The playbook that handles this alert is expected to take the
+# name of the down from the `node` variable, so we need to translate the
```
moolitayer: s/name of the down/name of the node/

cmd/autoheal/alerts_worker.go
```diff
@@ -120,94 +128,74 @@ func (h *Healer) cancelHealing(alert *alertmanager.Alert) error {
...
-	} else {
-		glog.Warningf(
-			"Healing rule '%s' has no actions, will have no effect on alert '%s'",
-			rule.ObjectMeta.Name,
-			alert.Name(),
-		)
 	}
-
-	return nil
+	matches = true
```
moolitayer: shouldn't matches be false in this case?

cmd/autoheal/healer.go
```diff
@@ -80,38 +73,44 @@ func NewHealerBuilder() *HealerBuilder {
...
+func (b *HealerBuilder) KubernetesClient(client kubernetes.Interface) *HealerBuilder {
+	b.k8sClient = client
 	return b
 }
 
 // Build creates the healer using the configuration stored in the builder.
 //
 func (b *HealerBuilder) Build() (h *Healer, err error) {
+	// Load the configuration files:
+	if b.configFiles == nil || len(b.configFiles) == 0 {
```
moolitayer: can be simplified to `len(b.configFiles) == 0` since [len of a nil object is 0](https://golang.org/pkg/builtin/#len)

cmd/autoheal/healer.go
```diff
@@ -167,29 +129,26 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
...
-	}
-
 	// Start the workers:
 	go wait.Until(h.runRulesWorker, time.Second, stopCh)
 	go wait.Until(h.runAlertsWorker, time.Second, stopCh)
 	glog.Info("Workers started")
 
+	// For each rule inside the configuration create a change and add it to the queue:
+	rules := h.config.Rules()
+	if rules != nil && len(rules) > 0 {
```
moolitayer: same

cmd/autoheal/healer.go
jhernand: Done.

cmd/autoheal/healer.go
jhernand: Done.

autoheal.yml
jhernand: Done.

cmd/autoheal/alerts_worker.go
jhernand: I think it should be `true`. The idea is that the method exists (returning `false`) as soon as it detects that the alert doesn't match. If it reaches this line then it means that the alert matched. In particular if the rule is empty then it will match all alerts, regardless of the labels of annotations that they have.

However, this method doesn't work correctly in some cases. In the next version of the pull request I am improving it and adding some tests to make sure that it works as expected.

jhernand:
The new version of the pull requests addresses the remaining comments. It also introduces a set of tests to verify that matching rules work correctly.

Not that in order to be able to run the tests with `make check` I also had to introduce a new commit that removes the `*_test.go` files from the embedded `junitreport` tool, as otherwise `make check` tries also to run those tests, and they fail because they can't be compiled with the `GOPATH` of the project. That may be an issue of the `release` project.

ironcladlou:
@jhernand 
>Not that in order to be able to run the tests with make check I also had to introduce a new commit that removes the *_test.go files from the embedded junitreport tool, as otherwise make check tries also to run those tests, and they fail because they can't be compiled with the GOPATH of the project. That may be an issue of the release project.

I think this will solve your problem: http://post-office.corp.redhat.com/archives/aos-devel/2018-February/msg00245.html

jhernand:
@ironcladlou not sure if that will actually solve the problem, as I don't initially wanted to exclude those packages, just to avoid building the tests. I will check, it may be good anyhow, to reduce the number of vendored files.

ironcladlou:
I think you **do** want to exclude `tools` from your project's package list, as that codebase is not actually a part of your project. Everything that things in `tools` is used for in the context of Make should work without you explicitly managing compilation of source within tools. It should be a self-contained and "benign" presence in your tree. Your code should have no actual Go dependency on anything from `tools`, otherwise all our components would be locked into the transitive dependencies of an unrelated project (tools).

I really don't like the way the tools thing is implemented.

jhernand:
Sure, I do want to exclude the `tools` dependencies, and I will try to use your suggestion for that. But that doesn't resolve my initial problem. This is the error that I see if I don't remove the `_test.go` files from the `tools` directory:

```
tools/junitreport/pkg/builder/flat/test_suites_builder.go:4:2: cannot find package "github.com/openshift/origin/tools/junitreport/pkg/api" in any of:
	/files/projects/openshift/autoheal/_output/local/go/src/github.com/openshift/autoheal/vendor/github.com/openshift/origin/tools/junitreport/pkg/api (vendor tree)
	/usr/lib/golang/src/github.com/openshift/origin/tools/junitreport/pkg/api (from $GOROOT)
	/files/projects/openshift/autoheal/_output/local/go/src/github.com/openshift/origin/tools/junitreport/pkg/api (from $GOPATH)
```

jhernand:
Note that I get that when running `make check`.

ironcladlou:
Adding the ignore setting to Gopkg.toml is supposed to solve exactly that sort of problem. Will try your branch locally...

jhernand:
Actually, this is what happens to my dependencies when I apply your suggestion of using the `ignore` parameter inside `Gopkg.toml` and I run `dep ensure -v`:

```
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	new file:   vendor/github.com/openshift/origin/LICENSE
	new file:   vendor/github.com/openshift/origin/cmd/cluster-capacity/go/src/github.com/kubernetes-incubator/cluster-capacity/LICENSE
	new file:   vendor/github.com/openshift/origin/cmd/service-catalog/go/src/github.com/kubernetes-incubator/service-catalog/LICENSE
	new file:   vendor/github.com/openshift/origin/hack/build-rpm-release.sh
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/api/string.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/api/test_case.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/api/test_suite.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/api/types.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/builder/flat/test_suites_builder.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/builder/interfaces.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/builder/nested/test_suites_builder.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/cmd/junitreport.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/cmd/summarize.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/parser/gotest/data_parser.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/parser/gotest/parser.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/parser/interfaces.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/parser/oscmd/data_parser.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/parser/oscmd/parser.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/parser/stack/interfaces.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/parser/stack/parser.go
	new file:   vendor/github.com/openshift/origin/tools/junitreport/pkg/parser/stack/stack.go

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   Gopkg.lock
	modified:   Gopkg.toml
	deleted:    vendor/github.com/davecgh/go-spew/LICENSE
	deleted:    vendor/github.com/davecgh/go-spew/spew/bypass.go
	deleted:    vendor/github.com/davecgh/go-spew/spew/bypasssafe.go
	deleted:    vendor/github.com/davecgh/go-spew/spew/common.go
	deleted:    vendor/github.com/davecgh/go-spew/spew/config.go
	deleted:    vendor/github.com/davecgh/go-spew/spew/doc.go
	deleted:    vendor/github.com/davecgh/go-spew/spew/dump.go
	deleted:    vendor/github.com/davecgh/go-spew/spew/format.go
	deleted:    vendor/github.com/davecgh/go-spew/spew/spew.go
	deleted:    vendor/github.com/hashicorp/golang-lru/2q.go
	deleted:    vendor/github.com/hashicorp/golang-lru/LICENSE
	deleted:    vendor/github.com/hashicorp/golang-lru/arc.go
	deleted:    vendor/github.com/hashicorp/golang-lru/doc.go
	deleted:    vendor/github.com/hashicorp/golang-lru/lru.go
	deleted:    vendor/github.com/hashicorp/golang-lru/simplelru/lru.go
	deleted:    vendor/github.com/hashicorp/golang-lru/simplelru/lru_interface.go
	deleted:    vendor/golang.org/x/net/context/context.go
	deleted:    vendor/golang.org/x/net/context/go17.go
	deleted:    vendor/golang.org/x/net/context/go19.go
	deleted:    vendor/golang.org/x/net/context/pre_go17.go
	deleted:    vendor/golang.org/x/net/context/pre_go19.go
	deleted:    vendor/k8s.io/apimachinery/pkg/apis/meta/internalversion/conversion.go
	deleted:    vendor/k8s.io/apimachinery/pkg/apis/meta/internalversion/doc.go
	deleted:    vendor/k8s.io/apimachinery/pkg/apis/meta/internalversion/register.go
	deleted:    vendor/k8s.io/apimachinery/pkg/apis/meta/internalversion/types.go
	deleted:    vendor/k8s.io/apimachinery/pkg/apis/meta/internalversion/zz_generated.conversion.go
	deleted:    vendor/k8s.io/apimachinery/pkg/apis/meta/internalversion/zz_generated.deepcopy.go
	deleted:    vendor/k8s.io/apimachinery/pkg/util/cache/cache.go
	deleted:    vendor/k8s.io/apimachinery/pkg/util/cache/lruexpirecache.go
	deleted:    vendor/k8s.io/apimachinery/pkg/util/diff/diff.go
	deleted:    vendor/k8s.io/client-go/discovery/fake/discovery.go
	deleted:    vendor/k8s.io/client-go/testing/actions.go
	deleted:    vendor/k8s.io/client-go/testing/fake.go
	deleted:    vendor/k8s.io/client-go/testing/fixture.go
	deleted:    vendor/k8s.io/client-go/tools/cache/controller.go
	deleted:    vendor/k8s.io/client-go/tools/cache/delta_fifo.go
	deleted:    vendor/k8s.io/client-go/tools/cache/doc.go
	deleted:    vendor/k8s.io/client-go/tools/cache/expiration_cache.go
	deleted:    vendor/k8s.io/client-go/tools/cache/expiration_cache_fakes.go
	deleted:    vendor/k8s.io/client-go/tools/cache/fake_custom_store.go
	deleted:    vendor/k8s.io/client-go/tools/cache/fifo.go
	deleted:    vendor/k8s.io/client-go/tools/cache/heap.go
	deleted:    vendor/k8s.io/client-go/tools/cache/index.go
	deleted:    vendor/k8s.io/client-go/tools/cache/listers.go
	deleted:    vendor/k8s.io/client-go/tools/cache/listwatch.go
	deleted:    vendor/k8s.io/client-go/tools/cache/mutation_cache.go
	deleted:    vendor/k8s.io/client-go/tools/cache/mutation_detector.go
	deleted:    vendor/k8s.io/client-go/tools/cache/reflector.go
	deleted:    vendor/k8s.io/client-go/tools/cache/reflector_metrics.go
	deleted:    vendor/k8s.io/client-go/tools/cache/shared_informer.go
	deleted:    vendor/k8s.io/client-go/tools/cache/store.go
	deleted:    vendor/k8s.io/client-go/tools/cache/thread_safe_store.go
	deleted:    vendor/k8s.io/client-go/tools/cache/undelta_store.go
	deleted:    vendor/k8s.io/client-go/tools/pager/pager.go
	deleted:    vendor/k8s.io/client-go/util/buffer/ring_growing.go
```

So actually, when using your suggestion, the `junitreport` package and its dependencies where *added* to the `vendor` directory. This probably means that I am doing something wrong, or that I misunderstood your comment.

ironcladlou:
I'm seeing even different behavior in my local clone... looks like we get to do another round of investigation into the current state of integrating the tools thing.

jhernand:
@ironcladlou `make check` after cleaning my environment, applying your suggestion, running `dep ensure -v` and restoring the tests of the `junitreport` project. So your suggestion has a positive efect, but I think it is because it *adds* the dependencies of those projects, not because it removes them.

Also, note that if I keep the `*_test.go` files of the `tools` directory, then those tests are executed when running `make check`:

```
ok  	github.com/openshift/autoheal/cmd/autoheal	1.452s	coverage: 11.8% of statements
ok  	github.com/openshift/autoheal/tools/junitreport/pkg/builder/flat	1.009s	coverage: 100.0% of statements
ok  	github.com/openshift/autoheal/tools/junitreport/pkg/builder/nested	1.008s	coverage: 98.5% of statements
ok  	github.com/openshift/autoheal/tools/junitreport/pkg/parser/oscmd	1.103s	coverage: 100.0% of statements
ok  	github.com/openshift/autoheal/tools/junitreport/pkg/parser/stack	1.009s	coverage: 17.2% of statements
```

Not a big problem, but it adds a bit of noise that I'd prefer to avoid.

ironcladlou:
Did you put the `ignored` statement at the top (i.e. before any `constraint`s)? That's important. If the ignore is effective, `tools` should be ignored complete by `dep`.

I'll spent some more time on this tomorrow. We may have to experiment with `required` or something. It's a mess.

jhernand:
@ironcladlou I think that this problem with the `tools` directory deservers its own issue, otherwise it will block this pull request for a long time. I have opened #15 to track it. If you don't have a major objection I will merge this pull request as is, then we can fix the problem in a better way in another pull request.

cmd/autoheal/alerts_worker.go
```diff
@@ -120,94 +128,77 @@ func (h *Healer) cancelHealing(alert *alertmanager.Alert) error {
...
-			"Running actions of healing rule '%s' for alert '%s'",
-			rule.ObjectMeta.Name,
-			alert.Name(),
-		)
-		for i := 0; i < len(rule.Spec.Actions); i++ {
-			err := h.runAction(rule, &rule.Spec.Actions[i], alert)
-			if err != nil {
-				return err
+func (h *Healer) checkMap(values, patterns map[string]string) (result bool, err error) {
+	if len(patterns) > 0 {
```
moolitayer: nit: I would prefer if result was explicitly set to false at this point for readability (if `len(patterns) > 0`)

That might be a personal preference though

jhernand: Sure, I am doing that change, then I am merging.

jhernand: Ah, you already merged the pull request, thanks :-) !

PR #11, opened 2018-03-26
Remove vendored dependencies

jhernand:
This pull request changes the `Makefile` of the project so that
the `vendor` directory will be populated as part of the execution
of the `build` target.

This way the more than 5000 files that are in the `vendor` directory
can be removed, thus reducing the size and review complexity of
future commits.


jhernand:
@ironcladlou, @moolitayer initially I commited the `vendor` directory, but after just a few commits and pull requests it is already clear that it isn't practical, so I think we should remove it and download the dependencies during the build with `dep --vendor-only`. Do you see any potential problem with that?

ironcladlou:
Standard practice in our community is to commit the vendor directory. I could give my own personal spiel about why I think that's a good standard for now, but it might be better to take the question to the aos-devel list if you want a more in depth discussion

ironcladlou:
FWIW churn in `vendor` should slow down very quickly as the project gets more code. I would recommend continuing to isolate vendor/dep changes to their own commits just for the sanity of reviewers

jhernand:
OK @ironcladlou, you have convinced me, will forget this for now.

moolitayer:
> I could give my own personal spiel about why I think that's a good standard for now

One drawback is having non repeatable builds: integration would not only be testing changes but also arbitrary master changes in dependencies. The same for version releases.

The tooling will get better eventually... :crossed_fingers:

jhernand:
Note that if you have a `Gopkg.lock` file and you use `dep ensure -vendor-only` to download the dependencies it will download always exactly the same versions, no matter what has changed in the repositories of those dependencies. The `Gopkg.lock` that is generated when you run `dep ensure` stores the exact commit that was selected, for example:

```toml
[[projects]]
  name = "github.com/emicklei/go-restful"
  packages = [
    ".",
    "log"
  ]
  revision = "2dd44038f0b95ae693b266c5f87593b5d2fdd78d"
  version = "v2.5.0"
```

Later, when you run `dep ensure -vendor-only` it downloads exactly that commit, and only for the dependencies that are already listed in the `Gopkg.lock` file. So in this sense the build is repeatable.

moolitayer:
And the lock is part of the source code. I see, thanks for clarifying that.

ironcladlou:
>Note that if you have a Gopkg.lock file and you use dep ensure -vendor-only to download the dependencies it will download always exactly the same versions, no matter what has changed in the repositories of those dependencies.

Assuming that the host of the source is actually available at all points in the future, and that the referenced tag is actually immutable, etc etc etc. Lots of repeatability concerns. In addition for CI specifically, it introduces a web of internet connectivity dependencies on external hosts for every build (which can be somewhat mitigated by caching proxies, like in Java/Maven world. No standards like this in Go land).

Lots of gotchas for sure.

PR #12, opened 2018-03-26
Prune dependencies

jhernand:
Version 0.3.2 of `dep` introuced a new pruning capability that reduces
considerable the number of files inside the `vendor` directory. That is
disabled by default, but can be enabled adding the following to the
`Gopkg.toml` file:

```toml
[prune]
  go-tests = true
  non-go = true
  unused-packages = true
```

This pull request uses pruning capability to reduce the number of vendored
files, from 5580 to 1270.


jhernand:
@moolitayer, @ironcladlou I think this is a good compromise. Dependencies are still commited, but the number of files is significantly smaller.

ironcladlou:
Great idea

PR #13, opened 2018-03-27
Use spf13/cobra to manage commad line

### LINKED ISSUE #7: Use spf13/cobra to manage command line
Currently the project uses the Go `flag` package to the command line
arguments, but it should use `spf13/cobra` like other OpenShift projects.

jhernand:
This patch changes the auto-heal binary so that it uses the
`spf13/cobra` to handle the command line. The binary will now have a
single `server` subcommand which starts the auto-heal service. For
example, a typical command line will be the following:
    
```
$ autoheal server --config-file=my.yml --logtostderr
```

Fixes #7.


jhernand:
@zgalor please review.

Please note that currently this pull request is based on #6, and also that it includes a patch to update the dependencies to include `spf13/cobra`, so please focus your review in the last commit, as it is the only reviewable/meaningful for this pull request.

zgalor:
@jhernand did you mean to tag @nimrodshn ?

jhernand:
@zgalor I added you because you commented that we should use `--config-file` instead of `-config-file`, and then @ironcladlou commented that we should use `spf13/cobra`.

@nimrodshn you are also welcome to review.

nimrodshn:
@jhernand Iooks much cleaner now :+1: 
Assuming other functionality is OK I approve these changes.

cmd/autoheal/server.go
```diff
@@ -0,0 +1,123 @@
...
+func serverRun(cmd *cobra.Command, args []string) {
+	var err error
+
+	// Set up signals so we handle the first shutdown signal gracefully:
+	stopCh := signals.SetupSignalHandler()
+
+	// Load the Kubernetes configuration:
+	var config *rest.Config
+	_, err = os.Stat(serverKubeConfig)
+	if os.IsNotExist(err) {
```
zgalor: What if there was another error?

jhernand: You are right @zgalor, that isn't checking all the potential errors. However, I think that it should be fixed in a separate patch. I created #14 to track it.

jhernand:
The new version of the pull requests is just a rebase, I will merge it now.

Issue #15, opened 2018-03-28
Avoid compiling and running the tests of `tools`


jhernand:
The `tools` directory of the project contains the source of some tools that have been copied from the `openshift/release` project, as decribed [here](https://github.com/openshift/release/tree/master/tools/hack/golang). A side effect of that is that when running `make check` the build scripts try to compile also the tests of those tools, in particular the tests of the `junitreport` tool. That fails because of missing dependencies:

```
tools/junitreport/pkg/builder/flat/test_suites_builder.go:4:2: cannot find package "github.com/openshift/origin/tools/junitreport/pkg/api" in any of:
	/files/projects/openshift/autoheal/_output/local/go/src/github.com/openshift/autoheal/vendor/github.com/openshift/origin/tools/junitreport/pkg/api (vendor tree)
	/usr/lib/golang/src/github.com/openshift/origin/tools/junitreport/pkg/api (from $GOROOT)
	/files/projects/openshift/autoheal/_output/local/go/src/github.com/openshift/origin/tools/junitreport/pkg/api (from $GOPATH)
```

Currently we solved this removing those tests files from the repository, but that isn't a good long term solution, as the tests may be accidentally re-added when the `tools` directory is updated in the future.

A possible solution suggested by @ironcladlou in #6 is to tell `dep` to ignore that directory, adding the following to the `Gopkg.toml` file:

```toml
ignored = ["github.com/openshift/autoheal/tools*"]
```

That seems like a more correct solution, but in our initial tests it didn't solve the problem. In addition, even if it solves the problem, then the tests of the `junitreport`  tool would be executed in addition to the tests of the project itself, thus polluting the output:

```
ok  	github.com/openshift/autoheal/cmd/autoheal	1.452s	coverage: 11.8% of statements
ok  	github.com/openshift/autoheal/tools/junitreport/pkg/builder/flat	1.009s	coverage: 100.0% of statements
ok  	github.com/openshift/autoheal/tools/junitreport/pkg/builder/nested	1.008s	coverage: 98.5% of statements
ok  	github.com/openshift/autoheal/tools/junitreport/pkg/parser/oscmd	1.103s	coverage: 100.0% of statements
ok  	github.com/openshift/autoheal/tools/junitreport/pkg/parser/stack	1.009s	coverage: 17.2% of statements
```

[...]

PR #16, opened 2018-03-28
[WIP] Remove tools dependencies

jhernand:
This pull request changes the `Gopkg.toml` file so that `dep` will not scan the
`tools` directory for dependencies. This has the positive effect of
removing from the `vendor` directory all the dependencies of those
tools, which aren't really dependencies of this project.


jhernand:
@ironcladlou, @moolitayer this is an attempt to address #15. Adding the `ignored` parameter to the `Gopkg.toml` file does make `dep` ignore the `tools` directory, so that the dependencies aren't added, as @ironcladlou suggested.

If I then revert the change that I used to remove the `*_test.go` files of the `tools` directory:

```
git revert 09e5d167ccdfe6636f4fd9280475d025a60f5fb3
```

Then `make works` correctly, but `make check` still tries to build those tests, and fails:

```
tools/junitreport/pkg/builder/flat/test_suites_builder.go:4:2: cannot find package "github.com/openshift/origin/tools/junitreport/pkg/api" in any of:
	/files/projects/openshift/autoheal/_output/local/go/src/github.com/openshift/autoheal/vendor/github.com/openshift/origin/tools/junitreport/pkg/api (vendor tree)
	/usr/lib/golang/src/github.com/openshift/origin/tools/junitreport/pkg/api (from $GOROOT)
	/files/projects/openshift/autoheal/_output/local/go/src/github.com/openshift/origin/tools/junitreport/pkg/api (from $GOPATH)
```

Any suggestion?

PR #19, opened 2018-03-28
Add alerts prefix to alerts endpoint

jhernand:
Currently the auto-heal service accepts alert notifications with any
request path, except with `metrics` because that is used to report
metrics to Prometheus. That could cause problems if in the future we
want to add other endpoints. To avoid that this patch changes the
service so that it will only accept alerts that use the `/alerts` prefix
in the request path.

Note that this means that in order to manually test you will have to
explicitly add the `/alerts` prefix:

```
$ autoheal server --config-file=my.yml --logtostderr
$ curl --data @manifests/node-down-alert.json http://localhost:9099/alerts
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,25 @@ awxJob:
     }
 ```
 
+## Testing
+
+To run the automated tests of the project run this command:
+
+```
+$ make check
+```
+
+To manually test the service, without having to have a running Prometheus alert
+manager that generates the alert notifications, you can use the `*-alert.json`
+files that are inside the `manifests` directory. For example, to simulate the
+`NodeDown` alert start the server and then use [curl](https://curl.haxx.se) to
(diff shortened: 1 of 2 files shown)


jhernand:
@zgalor @elad661 @moolitayer @ironcladlou please review.

README.md
```diff
@@ -184,6 +184,25 @@ awxJob:
...
+
+To run the automated tests of the project run this command:
+
+```
+$ make check
+```
+
+To manually test the service, without having to have a running Prometheus alert
+manager that generates the alert notifications, you can use the `*-alert.json`
+files that are inside the `manifests` directory. For example, so simulate the
```
moolitayer: s/so simulate/to simulate/

jhernand: Done.

PR #20, opened 2018-03-28
Adjust manifests for new configuration mechanism

jhernand:
This pull request adjusts the files in the `manifests` directory so that they
can be used to deploy the auto-heal service to an _OpenShift_ cluster,
after the recent changes in the command line and the configuration
mechanism.


jhernand:
@elad661, @zgalor, @moolitayer, @ironcladlou please review.

Note that the deployment of the auto-heal service should be implemented as an Ansible playbook in the `openshift-ansible` repository, but I think it is also good to have instructions and examples in this repository as well, so that it can be deployed manually, specially by developers.

manifests/README.md
```diff
@@ -0,0 +1,78 @@
...
+    ...
+```
+
+## Deployment
+
+Once you have customized the manifests as described in the above section, you
+can just run all of them using the `oc` or `kubectl` tools:
+
+```
+$ oc create -f *.yml
```
zgalor: maybe `oc apply` is better? it allows updating the configuration

manifests/autoheal-deployment.yml
```diff
@@ -29,6 +30,20 @@ spec:
...
+          name: autoheal
       containers:
       - name: autoheal
         image: openshift/origin-autoheal:latest
+        imagePullPolicy: IfNotPresent
+        volumeMounts:
+        - name: config
+          mountPath: /etc/autoheal
+        command:
+        - /usr/bin/autoheal
```
zgalor: isn't that the default entrypoint of the image?

manifests/autoheal-deployment.yml
```diff
@@ -29,6 +30,20 @@ spec:
...
+        args:
+        - server
```
zgalor: what is the `server` arg?

manifests/README.md
zgalor: also when you run with *.yml you depend on the alphabetical order of the files. So `autoheal-config-map.yml` will be created before the `autoheal-namespace.yml` is created, and it might fail...
Maybe we should add all the manifests to a single openshift template to eliminate dependencies?

manifests/autoheal-route.yml
```diff
@@ -19,7 +19,7 @@ kind: Route
 metadata:
   name: autoheal
 spec:
-  host: autoheal.private
+  host: autoheal.example.com
   port:
```
zgalor: I believe this section is not required. It is already stated in the service

manifests/autoheal-deployment.yml
moolitayer: We need to provide this arg now:
```
Usage:
  autoheal [command]

Available Commands:
  help        Help about any command
  server      Starts the auto-heal server
```

zgalor: In that case, I think we should make the server command default, and display help when `--help` flag is provided. Opened https://github.com/openshift/autoheal/issues/22 to track this

ironcladlou:
Related: we need a WIP PR like [this one](https://github.com/openshift/openshift-ansible/pull/7693) opened ASAP even if it's dependent on the user pushing their own images to the test cluster for development for now.

Would be nice if there is minimal duplication between whatever stuff you're adding here and the installer assets.

manifests/autoheal-route.yml
elad661: It's best not to specify the host in the Route object, so OpenShift will automatically assign one based on the service name, namespace, and the cluster's DNS settings.

manifests/README.md
```diff
@@ -0,0 +1,78 @@
...
+## Customization
+
+These manifests deploy the auto-heal service to a new `openshift-monitoring`
+namespace, which will be created by the `autoheal-namespace.yml` file. If you
+wish to use a different namespace then you will have to change all the files to
+reflect the new one. That can be done with the following command:
+
+```
+$ sed -i 's/openshift-monitoring/my-namespace/g' *.yml
+```
```
moolitayer: Openshift template?

jhernand:
The new version of the pull requests changes the approach, as suggested by the reviewers. It now replaces the multiple files inside the `manifests` directory with a single _OpenShift_ template, and a script showing how to use it.

moolitayer:
Looks good, I'm now attempting to run this against my [openshift cluster](https://github.com/openshift/release/tree/master/cluster/test-deploy) that is similar to what developers/ci use. Hoping I'll have relevant feedback after that

README.md
```diff
@@ -194,25 +208,18 @@ $ make check
...
-
-To build the RPM and the images, run this command:
-
-```
-$ make build-images
-```
+To install the service to an _OpenShift_ cluster use the template contained in
+the `template.yml` file. This template requires at the very minimum the address
+and the credentials to connect to the AWX or Ansible Tower server. See the
+`template.sh` script for an example of how to use it.  those parameters.
```
moolitayer: `those parameters.` seems leftover

template.yml
```diff
@@ -0,0 +1,185 @@
...
+metadata:
+  name: autoheal-template
+  annotations:
+    description: "Auto-heal service"
+    tags: "autoheal"
+
+parameters:
+- name: NAME
+  description: |-
+    The name of the auto-heal service. Will aso be used as a prefix for the
```
moolitayer: s/aso/also/

jhernand:
The new version of the pull request addresses the `aso` vs `also` typo.

PR #21, opened 2018-03-28
Return a meaningful error for missing template

moolitayer:
Currently if the template isn't found we fail silently since err is always nil at this point.


moolitayer:
@jhernand please review

jhernand:
Looks good, thanks @moolitayer.

PR #23, opened 2018-03-29
Return error if failed to obtain auth token

moolitayer:
In some cases, we fail to get the auth token, and continue to additional
requests that will fail. It is evident when using verbosity level 2:

```
connection.go:305] Sending POST request to 'http://localhost:9100/api/v2/authtoken/'.
connection.go:306] Request body:
{
  "username": "admin",
  "password": "password"
}
connection.go:330] Response body:
{
  "detail": "The requested resource could not be found."
}
connection.go:252] Sending GET request to 'http://localhost:9100/api/v2/job_templates/?name=Start+node&project__name=My+project'.
connection.go:274] Response body:
{
  "detail": "Authentication credentials were not provided."
}
```

Now we raise an error in those cases, and do not try any longer:
```
connection.go:305] Sending POST request to 'http://localhost:9100/api/v2/authtoken/'.
connection.go:306] Request body:
{
  "username": "admin",
  "password": "password"
}
connection.go:330] Response body:
{
  "detail": "The requested resource could not be found."
}
alerts_worker.go:62] Error obtaining auth token
```

### DIFF
--- a/pkg/awx/connection.go
+++ b/pkg/awx/connection.go
@@ -195,6 +195,9 @@ func (c *Connection) getToken() error {
 	if err != nil {
 		return err
 	}
+	if len(response.Token) == 0 {
+		return fmt.Errorf("Error obtaining auth token")
+	}
 	c.token = response.Token
 	return nil
 }


moolitayer:
@jhernand https://github.com/openshift/autoheal/pull/24 Is a more general solution so we might want to continue with that direction or keep both.

The reason I'm getting a 404 is that in my server (AWX 1.0.5.13) the auth endpoint is not as documented here:
http://docs.ansible.com/ansible-tower/2.3.0/html/towerapi/auth_token.html
But actually different:
My server has an additional unversioned oauth endpoint and does not have api/v2/authtoken/ or api/v1/authtoken/

```
{
    "oauth2": "/api/o/",
    "description": "AWX REST API",
    "available_versions": {
        "v1": "/api/v1/",
        "v2": "/api/v2/"
    },
    "custom_login_info": "",
    "current_version": "/api/v2/",
    "custom_logo": ""
}
```

That seems related to the recent https://github.com/ansible/awx/pull/904/files#diff-64d1ae443d41bb4ff887f175e05bc8d9L70

I thought the whole point in API versioning is to prevent that, meh

cc @elad661 @zgalor @ironcladlou

moolitayer:
> That seems related to the recent https://github.com/ansible/awx/pull/904/files#diff-64d1ae443d41bb4ff887f175e05bc8d9L70

So FYI
- we WILL have to change auth code at some point
- we might want to support both versions
- we need to figure out which tower version we are expected to work with.

moolitayer:
> @moolitayer can you please open a new issue for the authentication problem?

https://github.com/openshift/autoheal/issues/25

moolitayer:
@zgalor please review

PR #24, opened 2018-03-29
Return error when awx requests fail

moolitayer:
See https://github.com/openshift/autoheal/pull/23 for cases where we swallow errors.
This is a more general solution to that problem

```
Sending POST request to 'http://localhost:9100/api/v2/authtoken/'.
connection.go:313] Request body:
{
  "username": "admin",
  "password": "password"
}
alerts_worker.go:62] Status code '404' returned from server: '404 Not Found'
```

### DIFF
--- a/pkg/awx/connection.go
+++ b/pkg/awx/connection.go
@@ -260,6 +260,14 @@ func (c *Connection) rawGet(path string, query url.Values) (output []byte, err e
 	if err != nil {
 		return
 	}
+	if response.StatusCode != 200 {
+		err = fmt.Errorf(
+			"Status code '%d' returned from server: '%s'",
+			response.StatusCode,
+			response.Status,
+		)
+		return
+	}
 	body := response.Body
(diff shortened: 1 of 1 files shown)


pkg/awx/connection.go
```diff
@@ -260,6 +260,13 @@ func (c *Connection) rawGet(path string, query url.Values) (output []byte, err e
 	if err != nil {
 		return
 	}
+	if 300 <= response.StatusCode {
```
jhernand: Actually I think that we can be more strict here: anything other than 200 is unexpected and we don't currently know how to handle it, so:

```go
if resonse.StatusCode != 200 {
  ...
}
```

In the rest of the code, when the returned parameters have names, we are using just `return`:

```go
err = fmt.Errorf(...)
return
```

Can you do the same here?

moolitayer:
@jhernand addressed your comments, please take another look

Issue #25, opened 2018-04-01
Adjust autoheal to breaking auth API changes in AWX 3.3


moolitayer:
Currently we working with Ansible Tower 3.2. in version 3.3, the way we used to do authentication against the API server is dropped and we should adapt (either support only the new way or both).

See:
[Announcement - has links for several related docs](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/awx-project/kjlXPyc2YwE/5Zxi8PGFAgAJ;context-place=msg/awx-project/ei61pBqAT5I/FVjb4JzKCgAJ)
https://github.com/ansible/awx/pull/904

Originally mentioned in:
https://github.com/openshift/autoheal/pull/23

moolitayer:
I'm currently working on this issue. It will allow developing against AWX master (`devel`)

jhernand:
@moolitayer I'd suggest to first try to implement a simple solution: add support for specifying directly the token instead of the user name and password:

```go
// Get the token from the secret:
token := ...

// Create the connection:
connect, err := awx.ConnectionBuilder().
  Address(...).
  Proxy(...).
  Token(token).
  Build()
```

That token would then need to be created manually in the AWX UI. This is assuming that the token is long lived, I am not 100% sure about that.

rooftopcellist:
The token expiration time can be changed (currently only by manually changing the setting), but it defaults to 1000 years, which should be sufficiently long-lived.

jhernand:
Great, thanks for the input @rooftopcellist. Any chances you can take a look at the #26? That is the pull request where @moolitayer is implementing authentication using personal access tokens. No need to get into the details, just check that we are moving in the right direction.

rooftopcellist:
@jhernand The general approach of trying the AuthToken endpoint first and using the OAuth2 token if not should work.  For tower-cli (which also uses personal access tokens -- [here](https://github.com/ansible/tower-cli/pull/490/files#diff-6c200526bd6b171b824c81fc92924801R222)), we used a similar approach.  The only main difference afaik is that the existance of the `api/o` endpoint is checked when tower-cli starts and saved to determine which authentication method to use ([here](https://github.com/ansible/tower-cli/blob/9e57e0be580a73d94833c3932eed8694e2dbe4e8/tower_cli/utils/__init__.py#L39)).

jhernand:
[...]

PR #26, opened 2018-04-02
Adjust to new tower auth mechanism

moolitayer:
Fix https://github.com/openshift/autoheal/issues/25

Tower introduced auth changes in a non backward compatible manner, see issue.
We will keep trying the previous `api/v2/authtoken` endpoint.
If that fails we will attmept the new `api/v2/users/<name>/personal_tokens`
If that fails too, we will error out.

### DIFF
--- a/pkg/awx/connection.go
+++ b/pkg/awx/connection.go
@@ -41,6 +41,7 @@ type ConnectionBuilder struct {
 	password string
 	agent    string
 	token    string
+	bearer   string
 	insecure bool
 }
 
@@ -51,7 +52,9 @@ type Connection struct {
 	password string
 	agent    string
 	version  string
-	token    string
+	// AWX had two implementations for authentication tokens
+	token  string // using the /authtoken endpoint, used in tower < 3.3
+	bearer string // an OAuth2 implementation, used since tower 3.3
 
 	// The underlying HTTP client:
 	client *http.Client
@@ -101,6 +104,11 @@ func (b *ConnectionBuilder) Token(token string) *ConnectionBuilder {
 	return b
 }
 
+func (b *ConnectionBuilder) Bearer(bearer string) *ConnectionBuilder {
+	b.bearer = bearer
+	return b
+}
+
 func (b *ConnectionBuilder) Insecure(insecure bool) *ConnectionBuilder {
 	b.insecure = insecure
 	return b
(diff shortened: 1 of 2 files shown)


moolitayer:
@jhernand @zgalor @ironcladlou please take a look

pkg/awx/connection.go
```diff
@@ -263,7 +305,7 @@ func (c *Connection) rawGet(path string, query url.Values) (output []byte, err e
 	if err != nil {
 		return
 	}
-	if response.StatusCode != 200 {
+	if response.StatusCode > 202 {
```
moolitayer: @jhernand this is needed since after creating the new token the server returns the `created` 201 status code.

pkg/awx/connection.go
```diff
@@ -128,12 +136,14 @@ func (b *ConnectionBuilder) Build() (c *Connection, err error) {
...
+	authArgs := 0
+	for _, arg := range [3]string{b.username, b.token, b.bearer} {
+		if len(arg) > 0 {
+			authArgs++
+		}
 	}
-	if b.username == "" && b.token == "" {
-		err = fmt.Errorf("Either user name or token must be provided")
+	if authArgs != 1 {
+		err = fmt.Errorf("Exactly one of user name, token and bearer required")
```
zgalor: `Exactly one of the following is required: username, token or bearer`

pkg/awx/connection.go
```diff
@@ -178,7 +188,7 @@ func (c *Connection) Close() {
 // new onw to the server.
 //
 func (c *Connection) ensureToken() error {
-	if c.token != "" {
+	if c.token != "" || c.bearer != "" {
```
zgalor: I sometimes see the test `len(x) != 0` and sometime `!= ""`.
Is there a difference? do we want to follow a standard?

pkg/awx/connection.go
zgalor: don't we still want to check for other possible errors `!=200` ?
are codes 201-202 OK?
We might want to add a comment about that

pkg/awx/connection.go
```diff
@@ -326,7 +368,7 @@ func (c *Connection) rawPost(path string, query url.Values, input []byte) (outpu
 	if err != nil {
 		return
 	}
-	if response.StatusCode != 200 {
+	if response.StatusCode > 202 {
```
zgalor: same as previous comment

pkg/awx/connection.go
moolitayer: `len()` is more general (tests for nil as well) for string it's the same. I don't have a preference

pkg/awx/connection.go
moolitayer: 201 Created (we create tokens in this PR), 202 Accepted is what you sometimes get when you launch a background job. These (and 200 of course)are the most common in my experience.

pkg/awx/connection.go
moolitayer: Done

pkg/awx/connection.go
moolitayer: I'll change the other occurrence in this PR to be != as well

moolitayer:
Addressed comments @zgalor please take a look

jhernand:
@moolitayer I would like to merge this, can you rebase?

moolitayer:
@jhernand done

PR #28, opened 2018-04-02
Moving to viper as config file management

### LINKED ISSUE #8: Explore using spf13/viper to manage configuration
Currently the projects uses a custom configuration loader that reads
the YAML files directly. We should explore using the `spf13/viper` package
instead.

nimrodshn:
## What this does
This PR adds spf13/viper library as config file management to keep compatibility with other openshift products. fixes #8.

cc: @jhernand @moolitayer @zgalor

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -20,15 +20,14 @@ package config
 
 import (
 	"fmt"
-	"io/ioutil"
 
-	"github.com/ghodss/yaml"
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 
 	monitoring "github.com/openshift/autoheal/pkg/apis/monitoring/v1alpha1"
 	"github.com/openshift/autoheal/pkg/internal/data"
+	"github.com/spf13/viper"
 )
 
 // Loader contains the data and the methods needed to load the auto-heal service configuration.
@@ -105,16 +104,9 @@ func (l *Loader) Load() (config *Config, err error) {
 func (l *Loader) mergeFile(file string) error {
 	var err error
 
-	// Read the content of the file:
-	var content []byte
-	content, err = ioutil.ReadFile(file)
-	if err != nil {
-		return err
-	}
-
 	// Parse the YAML inside the file:
 	var decoded data.Config
-	err = yaml.Unmarshal(content, &decoded)
+	err = viper.Unmarshal(&decoded)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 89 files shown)


pkg/config/config.go
```diff
@@ -26,7 +26,7 @@ import (
 //
 type Config struct {
 	awx   *AWXConfig
-	rules []*monitoring.HealingRule
+	rules []monitoring.HealingRule
```
nimrodshn: Some changes had to be made to certain structs in order to do Unmarshalling.

pkg/config/loader.go
```diff
@@ -105,28 +104,21 @@ func (l *Loader) Load() (config *Config, err error) {
...
 	var decoded data.Config
-	err = yaml.Unmarshal(content, &decoded)
+	err = viper.Unmarshal(&decoded)
 	if err != nil {
 		return err
 	}
 
 	// Merge the configuration data from the file with the existing configuration:
-	if decoded.AWX != nil {
+	if &decoded.AWX != nil {
```
jhernand: I think this should be `decoded`, otherwise you are checking if the address of `decoded.AWX` is `nil`, and it will never be.

pkg/config/loader.go
```diff
@@ -105,28 +104,21 @@ func (l *Loader) Load() (config *Config, err error) {
...
 		err = l.mergeAWX(decoded.AWX)
 		if err != nil {
 			return err
 		}
 	}
-	if decoded.Rules != nil {
+	if &decoded.Rules != nil {
```
jhernand: Same, should be just `decoded.Rules`.

pkg/config/loader.go
```diff
@@ -136,7 +128,7 @@ func (l *Loader) mergeFile(file string) error {
 	return nil
 }
 
-func (l *Loader) mergeAWX(decoded *data.AWXConfig) error {
```
jhernand: `data.AWXConfig` is already a pointer, do we really need a pointer to a pointer?

pkg/internal/data/config.go
```diff
@@ -29,10 +29,10 @@ import (
 //
 type Config struct {
 	// AWX contains the details to connect to the default AWX server.
-	AWX *AWXConfig `json:"awx,omitempty"`
+	AWX AWXConfig `json:"awx,omitempty"`
```
jhernand: Does this mean that Viper doesn't know to work with pointers?

pkg/config/loader.go
jhernand: Sorry, that ^ was incorrect, please disregard.

nimrodshn:
@jhernand split into two commits.
Turnes out viper does know how to handle pointers -  I just never used it with pointers so I assumed it didn't..

nimrodshn:
@jhernand valid comments. I will explore these and make a patch for that. But this is a good starting point IMHO.

PR #31, opened 2018-04-02
Avoid `logging before flag.Parse`

jhernand:
That message is printed before every log message because `glog` thinks
that the command line hasn't been parsed:

```
ERROR: logging before flag.Parse: ...
```

The command line has actually been parsed, but by the `pflag` package,
not by the `flag` package. To avoid that, this pull request changes the main
program so that it parses a dummy empty command line, which makes `glog`
happy.

### DIFF
--- a/cmd/autoheal/main.go
+++ b/cmd/autoheal/main.go
@@ -37,6 +37,12 @@ func init() {
 }
 
 func main() {
+	// This is needed to make `glog` believe that the flags have already been parsed, otherwise
+	// every log messages is prefixed by an error message stating the the flags haven't been
+	// parsed.
+	flag.CommandLine.Parse([]string{})
+
+	// Execute the root command:
 	rootCmd.SetArgs(os.Args[1:])
 	rootCmd.Execute()
 }


jhernand:
@zgalor @moolitayer @elad661 please review.

nimrodshn:
Just to give some context - This issue and the workaround suggested are discussed extensively [here](https://github.com/kubernetes/kubernetes/issues/17162). No alternative solution that I have found, seems like kubernetes are [using the same solution as well](https://github.com/MrHohn/dns/blob/5745ce5b5c5871783ef11d2c19fe5b7fe9c1539f/cmd/kube-dns/dns.go#L43).

eparis:
FYI if you are using `AddGoFlagSet` I just merged https://github.com/spf13/pflag/pull/162 which should automatically apply this hack for you.

nimrodshn:
@eparis Nice! yes we are using `AddGoFlagSet` (see [here](https://github.com/openshift/autoheal/blob/924d5f8416c233d8c9e9daf4b7dad7518bc4e5cf/cmd/autoheal/main.go#L36)) - which renders this patch usless :sob:

PR #32, opened 2018-04-03
Healing action throttling

jhernand:
This patch adds a mechanism to throttle the execution of healing
actions. A new `throttling` section will be added to the configuration
file:

```yaml
throttling:
  interval: 1h
```

The new `interval` parameter controls for how long are executed healing
actions remembered. If the same healing action is triggered a second
time within this interval it won't be executed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,32 @@ should always be set to `false` (the default) in production environments.
 The `project` parameter is the name of the AWX project that contains the job
 templates that will be used to run the playbooks.
 
+### Throttling configuration
+
+The `throttling` section of the configuration describes how to throttle the
+execution of healing actions. This is intended to prevent _healing storms_ that
+could happen if the same alerts are send repeatedly to the service.
+
+The `interval` parameter controls the time that the service will remember an
+executed healing action. If an action is triggered more than once in the given
+interval it will be executed only the first time. The rest of the times it will
+be logged and ignored. (see `autoheal.yml` for an example.)
+
+The default interval value is one hour. Leaving the `interval` parameter 0
+will *disable* throttling altogether.
+
+Note that for throttling purposes actions are considered the same if they
+have exactly the same fields with exactly the same values *after* processing
+them as templates. For example, an action defined like this:
+
+```yaml
+awxJob:
+  template: "Restart {{ $labels.service }}"
+``
+
+Will have different values for the `template` field if the triggering alerts
+have different `service` labels.
+
 ### Healing rules configuration
 
 The second important section of the configuration file is `rules`. It contains
--- a/autoheal.yml
+++ b/autoheal.yml
@@ -63,6 +63,30 @@ awx:
   project: "My project"
 
 #
+# This section describes how to throttle the execution of healing rules.
+throttling:
+  #
+  # The time that the service will remember an executed healing action. If an
+  # action is triggered more than once in the given interval it will be executed
+  # only the first time. The rest of the times it will be logged and ignored.
+  #
+  # Note that for throttling purposes actions are considered the same if they
+  # have exactly the same fields with exactly the same values *after* processing
+  # them as templates.  For example, an action defined like this:
+  #
+  #   awxJob:
+  #     template: "Restart {{ $labels.service }}"
+  #
+  # Will have different values for the `template` field if the triggering alerts
+  # have different `service` labels.
+  #
+  # Interval is a duration string ,that is, a sequence of decimal numbers,
+  # each with optional fraction and a unit suffix, such as "300ms" or "2h45m".
+  # Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h". (For more details see https://golang.org/pkg/time/#Duration)
+  #
+  interval: 1h
+
+#
 # This section contains the healing rules.
 #
 rules:
--- a/cmd/autoheal/alerts_worker.go
+++ b/cmd/autoheal/alerts_worker.go
@@ -17,9 +17,11 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"regexp"
 
 	"github.com/golang/glog"
+	batch "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/util/runtime"
 
 	alertmanager "github.com/openshift/autoheal/pkg/alertmanager"
@@ -174,7 +176,23 @@ func (h *Healer) runRule(rule *monitoring.HealingRule, alert *alertmanager.Alert
 		alert.Name(),
 	)
 
-	// Prepare the template to process the action:
+	// Make a copy of the action so that we can modify it without affecting the rule stored in the
+	// cache:
+	var action interface{}
+	if rule.AWXJob != nil {
+		action = rule.AWXJob.DeepCopy()
+	} else if rule.BatchJob != nil {
+		action = rule.BatchJob.DeepCopy()
+	} else {
+		glog.Warningf(
+			"There are no action details, rule '%s' will have no effect on alert '%s'",
+			rule.ObjectMeta.Name,
+			alert.Name(),
+		)
+		return nil
+	}
+
+	// Process the templates inside the action:
 	template, err := NewObjectTemplateBuilder().
 		Variable("alert", ".").
 		Variable("labels", ".Labels").
@@ -183,28 +201,36 @@ func (h *Healer) runRule(rule *monitoring.HealingRule, alert *alertmanager.Alert
 	if err != nil {
 		return err
 	}
+	err = template.Process(action, alert)
+	if err != nil {
+		return err
+	}
 
-	// Decide which kind of action to run, and run it:
-	if rule.AWXJob != nil {
-		action := rule.AWXJob.DeepCopy()
-		err = template.Process(action, alert)
-		if err != nil {
-			return err
-		}
-		return h.runAWXJob(rule, action, alert)
-	} else if rule.BatchJob != nil {
-		action := rule.BatchJob.DeepCopy()
-		err = template.Process(action, alert)
-		if err != nil {
-			return err
-		}
-		return h.runBatchJob(rule, action, alert)
-	} else {
-		glog.Warningf(
-			"There are no action details, rule '%s' will have no effect on alert '%s'",
+	// Discard the action if it has been executed recently:
+	if h.actionMemory.Has(action) {
+		glog.Infof(
+			"Action for rule '%s' and alert '%s' has been executed recently, it will be ignored",
 			rule.ObjectMeta.Name,
 			alert.Name(),
 		)
+		return nil
 	}
-	return nil
+
+	// Execute the action:
+	switch typed := action.(type) {
+	case *monitoring.AWXJobAction:
+		err = h.runAWXJob(rule, typed, alert)
+	case *batch.Job:
+		err = h.runBatchJob(rule, typed, alert)
+	default:
+		err = fmt.Errorf(
+			"Don't know how to execute action of type '%T'",
+			typed,
+		)
+	}
+
+	// Remember that the action was executed recently, even if the execution failed:
+	h.actionMemory.Add(action)
+
+	return err
 }
--- a/cmd/autoheal/healer.go
+++ b/cmd/autoheal/healer.go
@@ -35,6 +35,7 @@ import (
 
 	"github.com/openshift/autoheal/pkg/alertmanager"
 	"github.com/openshift/autoheal/pkg/config"
+	"github.com/openshift/autoheal/pkg/memory"
 )
 
 // HealerBuilder is used to create new healers.
@@ -64,6 +65,9 @@ type Healer struct {
 	// notifications from the alert manager:
 	rulesQueue  workqueue.RateLimitingInterface
 	alertsQueue workqueue.RateLimitingInterface
+
+	// Executed actions will be stored here in order to prevent repeated execution.
+	actionMemory *memory.ShortTermMemory
 }
 
 // NewHealerBuilder creates a new builder for healers.
@@ -119,10 +123,19 @@ func (b *HealerBuilder) Build() (h *Healer, err error) {
 	glog.Infof("AWX user is '%s'", config.AWX().User())
 	glog.Infof("AWX project is '%s'", config.AWX().Project())
 
+	// Create the actions memory:
+	actionMemory, err := memory.NewShortTermMemoryBuilder().
+		Duration(config.Throttling().Interval()).
+		Build()
+	if err != nil {
+		return
+	}
+
 	// Allocate the healer:
 	h = new(Healer)
 	h.k8sClient = b.k8sClient
 	h.config = config
+	h.actionMemory = actionMemory
 
 	// Initialize the map of rules:
 	h.rulesCache = new(syncmap.Map)
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
@@ -87,9 +88,12 @@ func (l *Loader) Files(files []string) *Loader {
 // Load loads the configuration files and returns the resulting configuration object.
 //
 func (l *Loader) Load() (config *Config, err error) {
-	// Create an empty configuration:
+	// Create an default configuration:
 	l.config = &Config{
 		awx: &AWXConfig{},
+		throttling: &ThrottlingConfig{
+			interval: 1 * time.Hour,
+		},
 	}
 
 	// Merge the contents of the files into the empty configuration:
@@ -175,6 +179,12 @@ func (l *Loader) mergeFile(file string) error {
 			return err
 		}
 	}
+	if decoded.Throttling != nil {
+		err = l.mergeThrottling(decoded.Throttling)
+		if err != nil {
+			return err
+		}
+	}
 	if decoded.Rules != nil {
 		err = l.mergeRules(decoded.Rules)
 		if err != nil {
@@ -218,6 +228,17 @@ func (l *Loader) mergeAWX(decoded *data.AWXConfig) error {
 		l.config.awx.project = decoded.Project
 	}
 
+	return nil
+}
+
+func (l *Loader) mergeThrottling(decoded *data.ThrottlingConfig) error {
+	if decoded.Interval != "" {
+		interval, err := time.ParseDuration(decoded.Interval)
+		if err != nil {
+			return err
+		}
+		l.config.throttling.interval = interval
+	}
 	return nil
 }
 
--- a/pkg/memory/short_term.go
+++ b/pkg/memory/short_term.go
@@ -1,0 +1,159 @@
+/*
+Copyright (c) 2018 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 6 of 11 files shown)


jhernand:
@nimrodshn this is the WIP throttling change that I mentioned in another conversation.

nimrodshn:
@jhernand @moolitayer @zgalor  PTAL.

I have a number of notes on the current proposal that I wish to open for discussion:

1. `actionMemory` checks to purge items in two scenarios: (a) Uppon calling `Has`. (b) Every ten minutes it runs the purge method in a seperate worker - this number is up for discussion but I do think there is no need to keep these objects in memory until the next `Has` is called. Which leads to my second note -

2. Following a discussion with @moolitayer he noted that the items in `actionMemory` are a subset of the items in `rulesCache` and in fact we can add a timestamp field to the `HealingRule` struct and apply the same exact logic of `actionMemory` on the `rulesCache` - But, if there aren't a lot of rules to begin with than this implementation should be fine.

Either way I've tested this implementation and it works well with our current direction. (See tests)

jhernand:
I'd suggest you consider avoiding the periodical purging. You can instead purge only/exactly when it is necessary. At any point you know exactly when the next purge is needed, so instead of doing it periodically you can sleep till that time, and then do the purging.

The items in `actionMemory` aren't always the same that the items in the `rulesCache`. An action is actually a template, using the Go template language. For example, an action could be the following:

```yaml
awxJob:
  template: "Restart {{ $labels.service }}"
```

That would produce different action instances for different alerts, even if it is exactly the same rule. For example, if the alert is like this:

```yaml
labels:
  instance: ....
  service: httpd
```

It will produce the following action instance:

```yaml
awxJob:
  template: "Restart httpd"
```

But if the alert is like this:

```yaml
labels:
  instance: ...
  service: postgresqld
```

It will produce the following action instance:

```yaml
awxJob:
  template: "Restart postgresqld"
```

What we need to control is the action instances, not the templates. For example, it is OK to restart `httpd` and then inmediately restart `postgresql`. What is not OK is to restart `httpd` twice in a row.

So, I think that the timestamp should not be stored in the `rulesCache`.

moolitayer:
@nimrodshn my comment was different. I asked why do you need to purge past timers.

nimrodshn:
@moolitayer Oh sorry,  I misunderstood you then - @jhernand I like you're Idea! sounds good!

pkg/memory/short_term.go
```diff
@@ -0,0 +1,150 @@
...
+
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	return len(m.cells)
+}
+
+// PurgeExpiredCells finds the aged cells and removes them.
+//
+func (m *ShortTermMemory) PurgeExpiredCells() {
```
zgalor: suggestion: make this private lock-free, and wrap it with a public method that acquires the lock.
Then, from `Len()`, `Has()` and other methods, call the lock-free private `purge` from inside the lock context.
This will save lock acquiring time

pkg/memory/short_term.go
```diff
@@ -0,0 +1,150 @@
...
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	return m.findMatchingCell(item) != nil
+}
+
+// Len returns the number of items inside the memory.
+//
+func (m *ShortTermMemory) Len() int {
+	m.PurgeExpiredCells()
```
zgalor: why is this required here?
what is the usage `Len()` method?

README.md
```diff
@@ -100,6 +100,29 @@ data:
 The `project` parameter is the name of the AWX project that contains the job
 templates that will be used to run the playbooks.
 
+### Throttling configuration
+
+The `throttling` section of the configuration describes how to throttle the
+execution of healing actions. This is intended to prevent _healing storms_ that
+could happen if the same alerts are send repeatedly to the service.
+
+The `interval` parameter controls the time that the service will remember an
```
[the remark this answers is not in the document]
nimrodshn: Setting interval: 0s / 0h etc or leaving the value empty disables throttling altogether.

pkg/config/loader.go
```diff
@@ -82,9 +83,12 @@ func (l *Loader) Files(files []string) *Loader {
 // Load loads the configuration files and returns the resulting configuration object.
 //
 func (l *Loader) Load() (config *Config, err error) {
-	// Create an empty configuration:
+	// Create an default configuration:
 	l.config = &Config{
 		awx: &AWXConfig{},
+		throttling: &ThrottlingConfig{
+			interval: 1 * time.Hour,
```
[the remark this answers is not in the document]
nimrodshn: Documented.

pkg/memory/short_term.go
nimrodshn: IMHO we should purge cells on checking the length. @jhernand ?

pkg/memory/short_term.go
nimrodshn: I kind of like it this way :( 
@jhernand  what do you think?

nimrodshn:
@jhernand @zgalor 
The current behavior is this: 

1. If a user specifies `interval: someDuration` than the intial implementation is kept the same.

2. If a user specifies `interval: 0h,0s etc` or just leaves interval as an empty string than we disable throttling altogether. (without even creating the `actionMemory`. see [here](https://github.com/openshift/autoheal/pull/32/files#diff-3c6d47d7b38aeac558768f51b2845cadR125).)
3. If a user does not specify throttling at all we fall back to default `1h`.

All of this is documented.

nimrodshn:
@jhernand if throttling is enabled than we run a purge once every interval. (see [here](https://github.com/openshift/autoheal/pull/32/files#diff-3c6d47d7b38aeac558768f51b2845cadR160))

cmd/autoheal/healer.go
```diff
@@ -113,6 +122,20 @@ func (b *HealerBuilder) Build() (h *Healer, err error) {
...
+			Duration(config.Throttling().Interval()).
+			Build()
+
+		h.throttling_disabled = false
+		h.actionMemory = actionMemory
+	} else {
+		h.actionMemory = nil
+		h.throttling_disabled = true
+	}
+
```
nimrodshn: Note that this is possible due to the behavior of `time.ParseDuration` (see [here](https://play.golang.org/p/0cFkDyo8TCt))

autoheal.yml
```diff
@@ -62,6 +62,30 @@ awx:
...
+  # have exactly the same fields with exactly the same values *after* processing
+  # them as templates.  For example, an action defined like this:
+  #
+  #   awxJob:
+  #     template: "Restart {{ $labels.service }}"
+  #
+  # Will have different values for the `template` field if the triggering alerts
+  # have different `service` labels.
+  #
+  # interval is a duration string ,that is, a sequence of decimal numbers,
```
jhernand: Sorry, but I can't forget the typing classes I took when I was very young :-) . Can you start the sentence with upper case and wrap the lines at 80 characters?

cmd/autoheal/alerts_worker.go
```diff
@@ -183,28 +201,38 @@ func (h *Healer) runRule(rule *monitoring.HealingRule, alert *alertmanager.Alert
...
-		if err != nil {
-			return err
-		}
-		return h.runAWXJob(rule, action, alert)
-	} else if rule.BatchJob != nil {
-		action := rule.BatchJob.DeepCopy()
-		err = template.Process(action, alert)
-		if err != nil {
-			return err
+	if !h.throttling_disabled {
```
jhernand: If we get the same effect setting throttling to 0, then I think this isn't needed. If you still think that this is important, then consider adding a specific flag to enable/disable it:

```yaml
throttling:
  enabled: false
  interval: 1h
```

That way the user can temporarily disable it without having to modify the value of the `interval` parameter.

Also, please use camel case for variables: `throttlingDisabled`.

cmd/autoheal/healer.go
```diff
@@ -113,6 +122,20 @@ func (b *HealerBuilder) Build() (h *Healer, err error) {
 	h.k8sClient = b.k8sClient
 	h.config = config
 
+	// Create the actions memory:
+	// if Throttling is "" or 0h then ignore throttling altogether.
+	if config.Throttling().Interval() != 0*time.Second {
```
jhernand: You say it is "" or 0h, but it will never be "" as it isn't an string, at least not here.

Why `0*time.Second`? Isn't `0` enough?

Note also that I don't see the need for not creating the memory when throttling is disabled, it should work anyhow, and conditionally creating it or not complicates things here and in other places. I'd suggest to always create it, and make sure that it works as expected when the interval is zero.

cmd/autoheal/healer.go
```diff
@@ -133,6 +156,10 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
 	// Start the workers:
 	go wait.Until(h.runRulesWorker, time.Second, stopCh)
 	go wait.Until(h.runAlertsWorker, time.Second, stopCh)
+	// check for expired cells every interval.
+	if !h.throttling_disabled {
```
jhernand: I think that the fact that expired cells need to be purged should be hidden to the user of the memory object. If you need to do some cleanup when stopping then I'd suggest to add a `Close` method to the `Healer` type. That method should in turn call the `Close` method of the memory, which should then stop the goroutine that does the purging.

Note also that purging at intervals is a poor solution when you know exactly when you need to do the next purging. However it is acceptable in this case because it is combined with purging each time you call the `Has` method.

pkg/memory/short_term.go
jhernand: The `Len` method is needed for the tests, it is used to verify that expired cells have been actually removed when they should be removed. So, yes, we need to purge before calculating the length.

pkg/memory/short_term.go
jhernand: I guess that you made this method public in order to be able to call it from the main program. I think that isn't ideal, as the details of purging the memory should better be hidden behind the memory object. I'd suggest to make it private. The suggestion to make it lock-free also makes sense to me.

pkg/memory/short_term.go
```diff
@@ -0,0 +1,154 @@
...
+	for idx, cell := range m.cells {
+		age := now.Sub(cell.stamp)
+		// purge cells in place.
+		if age >= m.duration {
+			// zeroing the value of the cell so that it wouldn't be refrenced by the underlying array
+			// causing GO's grabage collector to collect the allocated memory.
+			cell = nil
+			m.cells = append(m.cells[:idx], m.cells[idx+1:]...)
+		}
+	}
```
jhernand: I think that we could copy the non-purged cells in the same array, and then resize it only once after the loop. But that is just a minor optimization, so this is OK for me.

cmd/autoheal/healer.go
nimrodshn: We can always create it and use the implementation as before the only issue is in that case we would add items to the memory and the purge them right away on the next `Has` / run of purge worker - this whole behavior is not needed, but if you think we can stand to bare carrying these items in memory than I will revert to my latest commit.

nimrodshn: As for you're first comment you are absolutely correct :) :+1:

autoheal.yml
nimrodshn: Sure!

pkg/memory/short_term.go
nimrodshn: This reads easier in my mind :)

cmd/autoheal/healer.go
jhernand: Yes, I think that disabling throttling is only interesting for development environments, and even in that case we will want to have throttling, maybe with a shorter (but not zero) interval.

cmd/autoheal/healer.go
nimrodshn: We can also lose the purging worker altogether if you think it's okay to keep action items in memory in between `Has` calls. Note that next purging is *per action* - if my interval is 1h and I added three actions to actionMemory: one at 10m one at 20m and one at 30m than purging the first action should only be valid from 70m onwards, the second at 80m onwards and the third and 90m onwards.. So there is no *exact* purge - the purge time is different between different actions. But this seems like the most simple solution to me.. Do you have any suggestions other than running purge periodically paired with purge on `Has`?

jhernand: I don't see a big problem with keeping action instances in memory. In the worst case it would waste some memory till the next action is triggered.

Exact purge is important, but we already have it because we do i tin the `Has` method. I think we don't need more than that.

If we wanted to implement exact purge without doing it in the `Has` method we could do the following:

1. Have a `nextPurge` field in the type. This should contain the exact time when the next purge needs to be done.

2. Have a `purgeCh` channel.

3. Have a gorutine that creates a timer with the value of the `nextPurge` field, and then waits for the timer to expire or for the next message in the `purgeCh` channel. The message in the channel can be anything, for example `interface{}`, it is just a mechanism to wake up that goroutine. When the message in the channel is `nil` the goroutine finishes.

4. In the methods that update the memory also update the `nextPurge` value, and send an empty (but not `nil`) message to the `purgeCn` channel.

5. When the memory is closed, send `nil` to the `purge` channel.

With that we would only purge exactly when it is needed. But this is probably overkill, I think we don't really need it, your current solution is enough.

nimrodshn:
@jhernand reverted back to the initial commit,
I made the `purge` method private behind a public method which acquires the lock. (see [here](https://github.com/openshift/autoheal/pull/32/files#diff-97cc7b6c061d29f1e8929eddff79e960R153)) And I removed the worker as suggested so now cleaning the actionMemory is called on `Has`.

pkg/memory/short_term.go
nimrodshn: This is done in place.

README.md
```diff
@@ -100,6 +100,31 @@ data:
...
+executed healing action. If an action is triggered more than once in the given
+interval it will be executed only the first time. The rest of the times it will
+be logged and ignored. (see `autoheal.yml` for an example.)
+
+The default interval value is one hour. Leaving the `interval` parameter 0 followed by a time unit suffix (i.e "0h" or "0s") will *disable* throttling altogether.
```
moolitayer: 0s == 0h == 0
better not to use units with 0 here or anywhere else

nimrodshn: Thanks :+1:

README.md
```diff
@@ -100,6 +100,31 @@ data:
...
+
+Note that for throttling purposes actions are considered the same if they
+have exactly the same fields with exactly the same values *after* processing
```
moolitayer: s/fields/labels/

pkg/memory/short_term.go
```diff
@@ -0,0 +1,157 @@
...
+	now := time.Now()
+	for idx, cell := range m.cells {
+		age := now.Sub(cell.stamp)
+		// purge cells in place.
+		if age >= m.duration {
+			// zeroing the value of the cell so that it wouldn't be refrenced by the underlying array
+			// causing GO's grabage collector to collect the allocated memory.
+			cell = nil
+			m.cells = append(m.cells[:idx], m.cells[idx+1:]...)
+		}
```
moolitayer: It seems the timestamps in `cells` are monotonously increasing
That means that you could break when age < duration

nimrodshn: you mean do this?
```
if age < m.duration {
  m.cells = m.cells[idx:]
}
```
Yeah this is true. But I like this implementation and since we don't *add* more cells but rather substract cells than the capacity of the underlying array is less or equal to the amount of cells we add in the `append` thus the complexity is **amortized constant** (see [here](https://groups.google.com/forum/#!msg/golang-nuts/o5rFNqd0VjA/HzJHwgl1y6MJ)).
Plus if you don't erase the cells (i.e `cell = nil`) the underlying struct remains the same size...

nimrodshn: This is not true you have 'nil out' all the cells before you call append otherwise memory explodes and you don't really do append meaning it is O(retired items) which can be O(len). You assume that the interval is small but it doesn't have to be..

nimrodshn: Oh you're first idea should work just fine! I thought you meant what I wrote..

nimrodshn:
@moolitayer fixed :+1:

pkg/memory/short_term.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	return nil
+}
+
+func (m *ShortTermMemory) Duration() time.Duration {
+	return m.duration
+}
+
+func (m *ShortTermMemory) CleanShortTermMemory() {
```
jhernand: I think this should be just `Clean`, and it should be documented, if possible.

pkg/memory/short_term.go
```diff
@@ -0,0 +1,159 @@
...
+		cell = new(ShortTermCell)
+		cell.item = item
+		m.cells = append(m.cells, cell)
+	}
+	cell.stamp = time.Now()
+}
+
+func (m *ShortTermMemory) Has(item interface{}) bool {
+	// Purge cells before checking.
+	m.CleanShortTermMemory()
```
jhernand: Should we call here the private lock-free method, after acquiring the mutex?

pkg/memory/short_term.go
```diff
@@ -0,0 +1,159 @@
...
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	return m.findMatchingCell(item) != nil
+}
+
+// Len returns the number of items inside the memory.
+//
+func (m *ShortTermMemory) Len() int {
+	m.CleanShortTermMemory()
```
jhernand: Same, should we call the private lock-free method instead?

pkg/memory/short_term.go
```diff
@@ -0,0 +1,159 @@
...
+func (m *ShortTermMemory) purgeExpiredCells() {
+	now := time.Now()
+	for idx, cell := range m.cells {
+		age := now.Sub(cell.stamp)
+		// purge cells in place.
+		if age >= m.duration {
+			// zeroing the value of the cell so that it wouldn't be refrenced by the underlying array
+			// causing GO's grabage collector to collect the allocated memory.
+			cell = nil
+			m.cells = append(m.cells[:idx], m.cells[idx+1:]...)
```
jhernand: I think this logic is correct. Maybe a bit hard to understand. Would be nice to have a comment explaining that cells are ordered from oldest to newest.

Also take into account that when you do `cell = nil` what you are clearing is the `cell` loop variable, but you aren't touching the underlying array. If you really want to clear the item in the array you will need to do `m.cells[idx] = nil`.

nimrodshn:
@jhernand Thanks! Done. :+1:

pkg/memory/short_term.go
```diff
@@ -0,0 +1,159 @@
...
+func (m *ShortTermMemory) Len() int {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	m.purgeExpiredCells()
+	return len(m.cells)
+}
+
+// PurgeExpiredCells finds the aged cells and removes them.
+// The cells in ShortTermMemory are monotonously increasing - from "older" to "younger"
+// thus, it suffices to check for until age < duration and then - break.
```
jhernand: When I suggested to comment this I meant a comment inside the body, it doesn't need (and shouldn't be) part of the public documentation as it is not a contract of this method.

nimrodshn:
@jhernand fixed! :+1:

jhernand:
Looks good to me, thanks @nimrodshn. Before merging, can you squash to a single commit without [WIP] in the subject?

pkg/memory/short_term.go
```diff
@@ -0,0 +1,159 @@
...
+// Len returns the number of items inside the memory.
+//
+func (m *ShortTermMemory) Len() int {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	m.purgeExpiredCells()
+	return len(m.cells)
+}
+
+// PurgeExpiredCells finds the aged cells and removes them.
```
jhernand: The comment should start with `// purgeExpiredCells ...`, to match the name of the method.

nimrodshn:
@jhernand Fixed!

nimrodshn:
@jhernand added.

jhernand:
@nimrodshn The subject of the commit still says [WIP]. Fix that and I will merge.

jhernand:
@nimrodshn also take a look at the commit message, it contains the complete history of the previous commit. Fix that, please.

nimrodshn:
@jhernand Done and Done :+1:

PR #33, opened 2018-04-03
Expose a metric on launched AWX jobs

moolitayer:
An example of the exposed metric after running one `start-node` job and two `SAY_HELLO` jobs:

```
# HELP autoheal_actions_initiated_total Number of initiated healing actions
# TYPE autoheal_actions_initiated_total counter
autoheal_actions_initiated_total{rule="say-hello",template="SAY_HELLO",type="awxJob"} 2
autoheal_actions_initiated_total{rule="start-node",template="Start node",type="awxJob"} 1
```

### DIFF
--- a/cmd/autoheal/awx_job.go
+++ b/cmd/autoheal/awx_job.go
@@ -82,7 +82,7 @@ func (h *Healer) runAWXJob(rule *monitoring.HealingRule, action *monitoring.AWXJ
 		alert.Name(),
 	)
 	for _, template := range templatesResponse.Results() {
-		err := h.launchAWXJob(connection, template, action)
+		err := h.launchAWXJob(connection, template, action, rule)
 		if err != nil {
 			return err
 		}
@@ -91,7 +91,12 @@ func (h *Healer) runAWXJob(rule *monitoring.HealingRule, action *monitoring.AWXJ
 	return nil
 }
 
-func (h *Healer) launchAWXJob(connection *awx.Connection, template *awx.JobTemplate, action *monitoring.AWXJobAction) error {
+func (h *Healer) launchAWXJob(
+	connection *awx.Connection,
+	template *awx.JobTemplate,
+	action *monitoring.AWXJobAction,
+	rule *monitoring.HealingRule,
+) error {
 	templateId := template.Id()
 	templateName := template.Name()
 	launchResource := connection.JobTemplates().Id(templateId).Launch()
@@ -105,5 +110,6 @@ func (h *Healer) launchAWXJob(connection *awx.Connection, template *awx.JobTempl
 		"Request to launch AWX job from template '%s' has been sent",
 		templateName,
 	)
+	h.incrementAwxActions(action, rule.ObjectMeta.Name)
 	return nil
 }
--- a/cmd/autoheal/metrics_exporter.go
+++ b/cmd/autoheal/metrics_exporter.go
@@ -1,0 +1,41 @@
+package main
+
+import (
+	"net/http"
+
+	monitoring "github.com/openshift/autoheal/pkg/apis/monitoring/v1alpha1"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+var (
+	actionsInitiated = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "autoheal_actions_initiated_total",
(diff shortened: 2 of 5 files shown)


moolitayer:
@jhernand please review

cc @zgalor @twiest @ironcladlou

cmd/autoheal/awx_job.go
```diff
@@ -102,5 +107,11 @@ func (h *Healer) launchAWXJob(connection *awx.Connection, template *awx.JobTempl
 		"Request to launch AWX job from template '%s' has been sent",
 		templateName,
 	)
+	healInitiated(
+		map[string]string{
+			"heal_type":     "awsJob",
```
jhernand: s/awsJob/awxJob/

In other parts of the code we use the terms "healing action" and "healing rule". To be consistent with that I'd suggest the following names for the labels:

```
action_type
template_name
rule_name
```

Or maybe even shorter:

```
type
template
rule
```

cmd/autoheal/metrics_exporter.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+var (
+	promHandler = promhttp.Handler()
+
+	healsInitiated = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "autoheal_heals_initiated_total",
```
jhernand: For consistency with the terminology used in other places I'd suggest to name the metric `autoheal_actions_initiated_total` and the help could be ` Number of initiated healing actions`.

cmd/autoheal/server.go
```diff
@@ -107,6 +107,9 @@ func serverRun(cmd *cobra.Command, args []string) {
 		glog.Fatalf("Error building Kubernets API client: %s", err.Error())
 	}
 
+	// Register exported metrics
```
jhernand: Hyper-nit: can you end the sentence with punctuation? `... metrics:`

cmd/autoheal/metrics_exporter.go
```diff
@@ -0,0 +1,26 @@
...
+		},
+		[]string{"heal_type", "template_name", "heal_name"},
+	)
+)
+
+func initExportedMetrics() {
+	prometheus.MustRegister(healsInitiated)
+}
+
+func healInitiated(labels map[string]string) {
```
jhernand: Can we pass here a pointer to the action object instead of a map? That way the names of the labels only need to be part of this file. Would be nice if we can also avoid repeating them, using constants, for example.

moolitayer: Done, I had to pass also the rule name

moolitayer:
@jhernand addressed comments

ironcladlou:
As a dogfooding exercise can you also add a metrics documentation file describing all metrics which are exposed?

ironcladlou:
e.g. https://github.com/coreos/etcd/blob/master/Documentation/metrics.md

moolitayer:
Added a `metrics` doc inspired by the etcd doc in a new documentation dir. Going forward we could use that dir to split out individual sections from the main README.md

documentation/metrics.md
```diff
@@ -0,0 +1,31 @@
...
+
+The naming of metrics follows the suggested [Prometheus best practices](http://prometheus.io/docs/practices/naming/). A metric name has an `autoheal`, `go` or `process` prefix as its namespace and a subsystem prefix.
+
+## autoheal namespace metrics
+
+The metrics under the `autoheal` prefix expose information related to heal operations undertaken by the server
+
+### actions
+
+These metrics describe the status of heal actions attempted by the server.
```
jhernand: s/heal actions/healing actions/

cmd/autoheal/metrics_exporter.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	monitoring "github.com/openshift/autoheal/pkg/apis/monitoring/v1alpha1"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+var (
+	promHandler = promhttp.Handler()
```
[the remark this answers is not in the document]
jhernand: Maybe `h.MetricsHandler()`?

moolitayer:
Rebased over #36 and verified. @jhernand ready?

PR #34, opened 2018-04-03
Secure connection by default between autoheal and AWX

elad661:
This commit adds support for setting the CA root cert in the AWX
client, and defaults to validating this cert.

If no root cert is specified, the system trust store will be used.

To retain the "insecure" option (the previous default),
for use for development, a new "insecure" option was added
to the configuration file

I haven't tested this all the way through as I don't have an autoheal/awx environment properly set up at the moment. Please consider that before merging.

cc: @jhernand @nimrodshn @moolitayer

### DIFF
--- a/cmd/autoheal/awx_job.go
+++ b/cmd/autoheal/awx_job.go
@@ -34,6 +34,8 @@ func (h *Healer) runAWXJob(rule *monitoring.HealingRule, action *monitoring.AWXJ
 	awxProxy := h.config.AWX().Proxy()
 	awxUser := h.config.AWX().User()
 	awxPassword := h.config.AWX().Password()
+	awxCA := h.config.AWX().CA()
+	awxInsecure := h.config.AWX().Insecure()
 
 	// Get the name of the AWX project name from the configuration:
 	awxProject := h.config.AWX().Project()
@@ -47,7 +49,8 @@ func (h *Healer) runAWXJob(rule *monitoring.HealingRule, action *monitoring.AWXJ
 		Proxy(awxProxy).
 		Username(awxUser).
 		Password(awxPassword).
-		Insecure(true).
+		CACertificates(awxCA).
+		Insecure(awxInsecure).
 		Build()
 	if err != nil {
 		return err
--- a/pkg/awx/connection.go
+++ b/pkg/awx/connection.go
@@ -19,6 +19,7 @@ package awx
 import (
 	"bytes"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -43,6 +44,10 @@ type ConnectionBuilder struct {
 	token    string
 	bearer   string
 	insecure bool
+
+	// If cacrt is specified, the client will expect the server certificate
+	// to be signed by this CA chain. If it isn't, the client will use the host trust store.
+	cacrt []byte
 }
 
 type Connection struct {
@@ -114,6 +119,11 @@ func (b *ConnectionBuilder) Insecure(insecure bool) *ConnectionBuilder {
 	return b
 }
 
+func (b *ConnectionBuilder) CACertificates(cacrt []byte) *ConnectionBuilder {
+	b.cacrt = cacrt
+	return b
+}
+
 func (b *ConnectionBuilder) Build() (c *Connection, err error) {
 	// Check the URL:
 	if b.url == "" {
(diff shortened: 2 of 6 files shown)


nimrodshn:
> I haven't tested this all the way through as I don't have an autoheal/awx environment properly set up at the moment. Please consider that before merging.

Maybe add WIP just to make it explicit not-to-merge.

cmd/autoheal/awx_job.go
```diff
@@ -47,7 +49,8 @@ func (h *Healer) runAWXJob(rule *monitoring.HealingRule, action *monitoring.AWXJ
 		Proxy(awxProxy).
 		Username(awxUser).
 		Password(awxPassword).
-		Insecure(true).
+		CACrt(awxCA).
```
jhernand: `Crt` is a bit cryptic for a public method name, in my opinion. Can we use something more readable? Maybe `CACertificates`?

pkg/awx/connection.go
```diff
@@ -42,6 +43,10 @@ type ConnectionBuilder struct {
 	agent    string
 	token    string
 	insecure bool
+
+	// If cacrt is specified, the client will expect the server certificate
+	// to be signed by this CA. If it isn't, the client will use the host trust store.
```
jhernand: Note that the server certificate doesn't need to be directly signed by this CA, it just needs to be verifiable using these CA certificates (there can be multiple certificates) as the roots. Not really important, just wanted to make sure that you understand the difference.

pkg/awx/connection.go
```diff
@@ -137,11 +147,27 @@ func (b *ConnectionBuilder) Build() (c *Connection, err error) {
 		return
 	}
 
+	if b.cacrt != nil && len(b.cacrt) > 0 && b.insecure {
+		err = fmt.Errorf("CA certificate and insecure are mutually exclusive")
+		return
+	}
+	var cert_store *x509.CertPool
```
jhernand: Variables are usually named using camel case: `certStore`. Can you change that?

pkg/awx/connection.go
```diff
@@ -137,11 +147,27 @@ func (b *ConnectionBuilder) Build() (c *Connection, err error) {
 		return
 	}
 
+	if b.cacrt != nil && len(b.cacrt) > 0 && b.insecure {
+		err = fmt.Errorf("CA certificate and insecure are mutually exclusive")
```
jhernand: CA certificate*s*, in plural, there may be multiple.

pkg/awx/connection.go
```diff
@@ -137,11 +147,27 @@ func (b *ConnectionBuilder) Build() (c *Connection, err error) {
 		return
 	}
 
+	if b.cacrt != nil && len(b.cacrt) > 0 && b.insecure {
```
jhernand: Here you can use just `len(b.cacrt)` because `len(nil) == 0` (I learned that from @moolitayer).

pkg/config/loader.go
```diff
@@ -153,6 +153,10 @@ func (l *Loader) mergeAWX(decoded *data.AWXConfig) error {
 		}
 	}
 
+	// Merge insecure setting
+	l.config.awx.insecure = decoded.Insecure
+	fmt.Println("%s", decoded.Insecure)
```
nimrodshn: @elad661 Should probably leave this out :wink:

pkg/config/loader.go
```diff
@@ -153,6 +153,10 @@ func (l *Loader) mergeAWX(decoded *data.AWXConfig) error {
 		}
 	}
 
+	// Merge insecure setting
```
jhernand: Nit: end the sentence with punctuation: `... setting:`

jhernand:
@elad661 I added the `WIP` label. Please remember to remove it when you consider that the pull request is ready to merge.

pkg/awx/connection.go
```diff
@@ -137,11 +147,27 @@ func (b *ConnectionBuilder) Build() (c *Connection, err error) {
...
+		err = fmt.Errorf("CA certificate and insecure are mutually exclusive")
+		return
+	}
+	var cert_store *x509.CertPool
+	if b.cacrt != nil && len(b.cacrt) == 0 {
+		cert_store, err = x509.SystemCertPool()
```
elad661: Question: Do we want to use the system certificate pool *and* `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` by default, or just the system pool without appending the kubernetes ca.crt?

jhernand: I think that only the system certificate pool, as the AWX server certificate won't probably be verifiable with the Kubernetes CA.

elad661:
I addressed the review feedback in a new commit (that should be squashed before this PR will be merged). I'll test this soon and remove the WIP tag when I verify it's working

elad661:
Tested, works, and ready for final review, squash & merge :)

@jhernand one note though, I do think we should consider appending the Kubernetes ca certificate to the trust store by default, because if the AWX server runs on the same cluster and the autoheal would connect to it via an OpenShift `Route`, it would be signed by that CA, and users won't have to copy that CA to the secret.

On the other hand, in cases like this people might setup the autoheal to communicate to AWX via the service directly without going through an external `Route` - so I'm not sure it would actually be beneficial on any real-world setup.

moolitayer:
Maybe add in the doc of config.go `func (c *AWXConfig) CA() []byte `:
`If not provided the system cert pool will be used`

elad661:
@moolitayer Done (and squashed).

jhernand:
If we ever want to add the Kubernetes CA certificates automatically it would be in a different place, not in this code, as it should be a reasonably generic AWX client and there may be AWX clients that know nothing about Kubernetes.

I think that this pull request is good, thanks @elad661, I am merging it.

PR #48, opened 2018-04-12
Send alert log only when using --v=2

jhernand:
Currently we dump the bodies of the requests received from the alert
manager by default. This makes the log too verbose. This patch changes
the healer so that it do this only when the log verbosity level is set
to at least 2.


jhernand:
@elad661 @zgalor @moolitayer @nimrodshn please review.

zgalor:
LGTM :+1: 

In general, I think we need to decide on a log level set (maybe there is a standard one already in openshift) and apply it to all our logs.

PR #49, opened 2018-04-12
Make secrets optional

jhernand:
Currently the auto-heal server needs to read the AWX crendentials and CA
certificates from Kubernetes secret. This complicates things in
development environments, because the secrets have to be created before
using the server. To simplify things, this patch adds support for
reading the credentials and CA certificates directly from the
configuration file. For example:

```yaml
awx:
  credentials:
    username: autoheal
    password: ...
  tls:
    caFile: my-ca.pem
```

Using the `credentialsRef` and `tlsRef` sections to reference Kubernetes
is still supported and the recommended configuration when running the
server in production environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,36 @@ data:
   password: ...
 ```
 
+Alternatively it is also possible to specify the user name and password directly
+inside the configuration file, using the `credentials` section. For example:
+
+```yaml
+credentials:
+  username: autoheal
+  password: ...
+```
+
+This is very convenient for development environments, but it is not recommended
+for production environments, as then the configuration file needs to be
+protected very carefully. For example, you can create a separate file for the
+credentials, give it restricted permissions, and then load it using the
+`--config-file` option twice:
+
+```
+$ echo > general.yml <<.
+awx:
+  address: https://myawx.example.com/api
+.
+$ echo > credentials.yml <<.
+credentials:
+  username: "autoheal"
+  password: "..."
+.
+$ chmod u=r,g=,o= credentials.yml
+$ autoheal server --config-file=general.yml --config-file=credentials.yml
+
+```
+
 The `tlsRef` parameter is a reference to the [Kubernetes
 secret](https://kubernetes.io/docs/concepts/configuration/secret) that contains
 the certificates used to connect to the AWX API. That secret should contain the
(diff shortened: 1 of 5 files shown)


jhernand:
@nimrodshn @moolitayer @zgalor @elad661 please review.

autoheal.yml
```diff
@@ -34,17 +34,41 @@ awx:
   proxy: http://my-proxy.example.com:3128
 
   #
-  # Reference to the Kubernetes secret that contains the user name and password
-  # to use to connect to the AWX API.
+  # The user name and password used to connect to the AWX API.
+  #
+  credentials:
```
zgalor: What happens if both sections (`credentials` and `credentialsRef`) exist? Which one takes over?

I think that developer-oriented configuration should commented out in this file
```
# credentials:
#    username: my-user
#    password: my-password
```

autoheal.yml
```diff
@@ -34,17 +34,41 @@ awx:
...
   #
   credentialsRef:
     namespace: my-namespace
     name: my-awx-credentials
 
   #
-  # Reference to the Kubernetes secret that contains the trusted CA certificates
-  # used to verify the TLS certificate presented by the AWX server. If not
-  # present or empty then the global system trusted certificates will be used.
+  # Trusted CA certificates used to verify the TLS certificate presented by the
```
zgalor: Same comment as above

README.md
```diff
@@ -63,10 +63,38 @@ The `proxy` parameter is optional, and it indicates what HTTP proxy should be
...
+
+```yaml
+credentials:
+  username: autoheal
+  password: ...
+```
+
+Note that if you store the credentials in the configuration file then it needs
+to be protected very carefully. For example, you can create a separate file for
+the the credentials, give it restricted permissions, and then load it using the
```
moolitayer: double word `the`.
I prefer to discourage users from using this in production as secrets are better suited.

autoheal.yml
jhernand: Done in the next version of the pull request.

README.md
jhernand: In the next version of the pull request I am explicitly saying what is the recommended mechanism for production environments.

README.md
```diff
@@ -79,6 +79,36 @@ data:
...
+inside the configuration file, using the `credentials` section. For example:
+
+```yaml
+credentials:
+  username: autoheal
+  password: ...
+```
+
+This is very convenient for development environments, but it is not recommended
+for production environments, as then then the configuration file needs to be
```
moolitayer: double `then`

jhernand: Done.

PR #50, opened 2018-04-13
Use `autoheal.openshift.io` API group

### LINKED ISSUE #42: Rename the API group from `monitoring` to `autoheal`
The auto-heal service uses the Kubernetes API infrastructure to define the `HealingRule` type. This type is inside a `monitoring.openshift.io` API group because originally we planned to have other types in that group unrelated to auto-heal. This does not make sense now, so we should rename that group to `autoheal.openshift.io`, or similar.

jhernand:
The initial prototypes of the auto-heal service used the
`monitoring.openshift.io` API group because it was intended to contain
also types to define Prometheus alerting rules. That has been changed
during the design of the service, and it only takes care now of healing
rules. To reflect that this patch changes the API group to
`autoheal.openshift.io`.

Fixes #42.


jhernand:
@zgalor @moolitayer @elad661 @nimrodshn please review.

pkg/apis/autoheal/v1alpha1/doc.go
```diff
@@ -16,5 +16,5 @@ limitations under the License.
 
 // +k8s:deepcopy-gen=package
 
-// +groupName=monitoring.openshift.io
-package v1alpha1 // import "github.com/openshift/autoheal/pkg/apis/monitoring/v1alpha1"
+// +groupName=autoheal.openshift.io
+package v1alpha1 // import "github.com/openshift/autoheal/pkg/apis/autoheal/v1alpha1"
```
zgalor: Since you are changing the api, don't you want to change the package version?
something like `v1alpha2`?

nimrodshn:
LGTM :+1:

pkg/apis/autoheal/v1alpha1/doc.go
jhernand: Done in the next version of the pull request.

jhernand:
The new version of the pull requests also changes the API version to `v1alpha2`, as suggested by @zgalor.

PR #52, opened 2018-04-16
Start collection of example playbooks

jhernand:
This patch starts a collection of example auto-heal playbooks. The
initial playbooks in this collection are intended to resolve CPU and
memory issues in OpenShift nodes that run as virtual machines managed by
oVirt.

### DIFF
--- a/examples/playbooks/increase_ovirt_node_cpus.yml
+++ b/examples/playbooks/increase_ovirt_node_cpus.yml
@@ -1,0 +1,110 @@
+#
+# Copyright (c) 2018 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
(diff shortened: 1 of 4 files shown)


jhernand:
@nimrodshn @zgalor @moolitayer @elad661 these are the playbooks that I used for the demo that I shared with you recently. Please review.

examples/playbooks/increase_ovirt_node_cpus.yml
```diff
@@ -0,0 +1,91 @@
...
+    # Maximum number of CPUs to assing to the VM.
+    max_cpus: 4
+    # Amount of CPUs to add in each increase.
+    inc_cpus: 1
+
+  tasks:
+
+  # Calculate the IP address from the `instance` input variable, and then do a
+  # reverse DNS lookup to find the FQDN, which in this environment is equal to
+  # the name of the virtual machine.
```
moolitayer: Worth to mention another requirement for this hostname == k8s node name (returned by the alert)

jhernand: Note that the k8s node name is not involved at all here. The requirements are that the `instance` label of the alert contains the IP address of the node, and that the name of the virtual machine in oVirt is equal to the FQDN. The first is guaranteed by the way that the node exporter works. The second isn't guaranteed at all. I will try to add both things to the comments.

jhernand: Yes, when using the node exporter the `instance` label is the IP address and port of the metrics endpoint of the node for example `192.168.100.6:9100`.

jhernand: Note that this for the metrics that were used in this example, other metrics, from other sources, may have different `instance` labels.

PR #55, opened 2018-04-16
Adding tests for config pkg

nimrodshn:
## What this does?
Adds tests for config pkg.

cc: @jhernand  @moolitayer

### DIFF
--- a/pkg/config/loader_test.go
+++ b/pkg/config/loader_test.go
@@ -1,0 +1,351 @@
+/*
+Copyright (c) 2018 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"io/ioutil"
+	"os"
+	"reflect"
+	"strings"
+	"testing"
+	"time"
+
+	autoheal "github.com/openshift/autoheal/pkg/apis/autoheal/v1alpha2"
+	batch "k8s.io/api/batch/v1"
+	core "k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestFiles(t *testing.T) {
+	l := NewLoader()
+
+	data0 := `
+      awx:
+        address: "http://test_address.com"`
+
+	data1 := `
+      awx:
+        proxy: "http://test-proxy.com:1234"`
+
+	tempDir, _ := ioutil.TempDir("", "test_dir")
+
+	file0, _ := ioutil.TempFile(tempDir, "test_file0")
+	file1, _ := ioutil.TempFile(tempDir, "test_file1")
+
+	file0.WriteString(data0)
+	file1.WriteString(data1)
+
+	defer file0.Close()
+	defer file1.Close()
+
+	defer os.RemoveAll(tempDir)
+
+	l.Files([]string{file0.Name(), file1.Name()})
+	cfg, err := l.Load()
+
+	expected := &Config{
+		awx: &AWXConfig{
+			address: "http://test_address.com",
+			proxy:   "http://test-proxy.com:1234",
+			jobStatusCheckInterval: 5 * time.Minute,
+		},
+		throttling: &ThrottlingConfig{
+			interval: 1 * time.Hour,
+		},
+	}
+
+	if err != nil {
+		t.Errorf("An error occured! %s", err)
+	}
+
+	if !reflect.DeepEqual(cfg, expected) {
+		t.Errorf("Expected %q but got %q", expected, cfg)
+	}
+}
+
+func TestLoadFile(t *testing.T) {
+	filename := "test_config"
+	file, _ := ioutil.TempFile("", filename)
+
+	defer file.Close()
+	defer os.Remove(file.Name())
+
+	configsTest := []struct {
+		configString string
+		expected     *Config
+	}{
+		{
+			configString: "",
+			expected: &Config{
+				awx: &AWXConfig{
+					jobStatusCheckInterval: time.Duration(5) * time.Minute,
+				},
+				throttling: &ThrottlingConfig{
+					interval: time.Duration(1) * time.Hour,
+				},
+			},
+		},
+		{
+			configString: `
+               awx:
+                 address: https://my-awx.example.com/api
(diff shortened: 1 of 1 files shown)


pkg/config/loader_test.go
```diff
@@ -0,0 +1,305 @@
...
+import (
+	"os"
+	"reflect"
+	"strconv"
+	"testing"
+	"time"
+
+	autoheal "github.com/openshift/autoheal/pkg/apis/autoheal/v1alpha2"
+	batch "k8s.io/api/batch/v1"
+	v1 "k8s.io/api/core/v1"
```
jhernand: I think that in other places we use `core` instead of `v1`, can you do the same?

pkg/config/loader_test.go
```diff
@@ -0,0 +1,305 @@
...
+func TestFile(t *testing.T) {
+	l := NewLoader()
+	l.File("test_file")
+	if len(l.files) != 1 {
+		t.Fail()
+	}
+
+	if l.files[0] != "test_file" {
+		t.Fail()
+	}
```
jhernand: I think it would be better if we don't depend on the private `files` field, otherwise this test will stop working if we change those internals. I'd suggest to either remove this kind of test, or else rewrite it in such a way that it only uses the public interface of the `Loader` type.

pkg/config/loader_test.go
```diff
@@ -0,0 +1,305 @@
...
+	l.Files([]string{"test_file0", "test_file1", "test_file2"})
+	if len(l.files) != 3 {
+		t.Fail()
+	}
+
+	for idx, file_name := range l.files {
+		if file_name != ("test_file" + strconv.Itoa(idx)) {
+			t.Fail()
+		}
+	}
```
jhernand: Same, try to avoid using the private fields. In this particular case you can do the same putting different parameters in different files, and then checking that all of them have been loaded. For example, `test_file0` can contain this:

```yaml
awx:
  address: https://my-awx.example.com
```

Then `test_file1` can contain this:

```yaml
awx:
```

Then you can load both files and check that both `awx.address` and `awx.proxy` are populated and that they have the expected values.

pkg/config/loader_test.go
```diff
@@ -0,0 +1,305 @@
...
+				awx: &AWXConfig{
+					jobStatusCheckInterval: time.Duration(5) * time.Minute,
+				},
+				throttling: &ThrottlingConfig{
+					interval: time.Duration(1) * time.Hour,
+				},
+			},
+		},
+		{
+			configString: `awx:
```
jhernand: Can we indent this text? I mean something like this:

```go
configString: `
    awx:
      ...
`
```

Maybe we could add a helper function that writes the YAML text to a file, removing from all lines the white space prefix of the first line.

pkg/config/loader_test.go
nimrodshn: This is actually `v1` in `loader.go` so I kept it the same for consistency of testing..

nimrodshn:
@jhernand I tried for the strings to look as nice as I could. I can try and work on their indentation a little more and write some helper function to erase the whitespace after writing the strings to file so they will be parsed as yaml correctly - Do you think that is necessary? Other comments are fixed :smile: :+1:

pkg/config/loader_test.go
```diff
@@ -0,0 +1,330 @@
+package config
```
jhernand: Remember to add the copyright header, like in all the other source files in the project.

pkg/config/loader_test.go
```diff
@@ -0,0 +1,330 @@
...
+)
+
+func TestFiles(t *testing.T) {
+	l := NewLoader()
+
+	filename0 := "test_file0.yaml"
+	filename1 := "test_file1.yaml"
+
+	data0 := `
+  awx:
```
jhernand: I think this can look even nicer if you indent the YAML text with tabs instead of spaces, that way it will align correctly wit the Go code when displayed using the standard 8 spaces per tab.

pkg/config/loader_test.go
```diff
@@ -0,0 +1,330 @@
...
+    address: "http://test_address.com"`
+
+	data1 := `
+  awx:
+    proxy: "http://test-proxy.com:1234"`
+
+	file0, _ := os.Create(filename0)
+	file1, _ := os.Create(filename1)
```
jhernand: Can we create these files in a temporary directory? Maybe using `ioutil.TempDir` or `ioutil.TempFile`. Otherwise, if we create them in the project directory, chances are that someone will add them to the repository by accident.

pkg/config/loader_test.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	if !reflect.DeepEqual(cfg, expected) {
+		t.Errorf("Expected %q but got %q", expected, cfg)
+	}
+}
+
+func TestLoadFile(t *testing.T) {
+	filename := "test_config.yml"
+	file, _ := os.Create(filename)
```
jhernand: Same, use a temporary directory or file.

pkg/config/loader_test.go
```diff
@@ -0,0 +1,330 @@
...
+
+		if !reflect.DeepEqual(cfg, test.expected) {
+			t.Errorf("Expected %q but got %q", test.expected, cfg)
+		}
+	}
+}
+
+func TestLoadDir(t *testing.T) {
+	filename := "test_config.yml"
+	file, _ := os.Create(filename)
```
jhernand: Use a temporary directory or file.

pkg/config/loader_test.go
```diff
@@ -0,0 +1,351 @@
...
+
+	data0 := `
+      awx:
+        address: "http://test_address.com"`
+
+	data1 := `
+      awx:
+        proxy: "http://test-proxy.com:1234"`
+
+	temp_dir, _ := ioutil.TempDir(".", "test_dir")
```
jhernand: The convention for variable is camel case: `tempDir`.

Consider using `""` instead of `"."`, that way the temporary directory is created in `/tmp` instead of in the current working directory.

pkg/config/loader_test.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	if !reflect.DeepEqual(cfg, expected) {
+		t.Errorf("Expected %q but got %q", expected, cfg)
+	}
+}
+
+func TestLoadFile(t *testing.T) {
+	filename := "test_config"
+	file, _ := ioutil.TempFile(".", filename)
```
jhernand: Use `""` instead of `"."`.

pkg/config/loader_test.go
```diff
@@ -0,0 +1,351 @@
...
+		if !reflect.DeepEqual(cfg, test.expected) {
+			t.Errorf("Expected %q but got %q", test.expected, cfg)
+		}
+	}
+}
+
+func TestLoadDir(t *testing.T) {
+	filename := "test_config"
+
+	dir, _ := ioutil.TempDir(".", "temp_dir")
```
jhernand: Use `""` instead of `"."`.

nimrodshn:
@jhernand Done, Thanks :+1:

PR #56, opened 2018-04-16
Awx support for projects

jhernand:
This pull request adds to the AWX client basic support for retrieving projects, and some examples that show how to use that support.

### DIFF
--- a/pkg/awx/project_resource.go
+++ b/pkg/awx/project_resource.go
@@ -14,8 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// This file contains the implementation of the resource that manages a specific job
-// template.
+// This file contains the implementation of the resource that manages a specific project.
 
 package awx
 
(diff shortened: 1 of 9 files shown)


jhernand:
@zgalor @moolitayer @elad661 @nimrodshn we don't really need it, but it may be useful if we eventually move the AWX client to a different repository, and also as an example of how to add support for other kinds of objects.

pkg/awx/project_resource.go
```diff
@@ -0,0 +1,68 @@
...
+	request.resource = &r.Resource
+	return request
+}
+
+type ProjectGetRequest struct {
+	Request
+}
+
+func (r *ProjectGetRequest) Send() (response *ProjectGetResponse, err error) {
+	data := new(data.ProjectGetResponse)
```
zgalor: `data` is a bit confusing with the name of the package. I suggest `output` - this is what we use for `JobTemplatesResource`, `JobTemplateLaunchResource`, `JobResource` and `JobsResource`.

If you agree, you might as well change `JobTemplatesResource`, which also uses `data`

jhernand: Yes, I agree, will do.

PR #58, opened 2018-04-18
Fix configuration loader tests

jhernand:
This pull request fixes some issues with the configuration loader tests, in particular the following:

- `gofmt -s` likes to simplify the array literals, removing the redundant type specifications.

- `govet` doesn't like the use of `%q` in string formats.

- Recent changes made the `ca` field of the configuration a byte buffer, so it can no longer be compared with the default value.

These issues where detecting running `make check`.

### DIFF
--- a/pkg/config/loader_test.go
+++ b/pkg/config/loader_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package config
 
 import (
+	"bytes"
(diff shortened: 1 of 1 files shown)


nimrodshn:
@jhernand Thanks a lot :+1: Sorry for this :(

Issue #53, opened 2018-04-16
Improve exposed metrics usability


moolitayer:
Currently we expose one counter called `autoheal_actions_initiated_total` with labels `[template,type and rule]` that is incremented each time an AWX job is successfully launched[1][2]

Current gaps:
- No reporting on the recently added job statuses.
- If a heal failed prior to launching the action it is not reported.
- Maybe we need to expose received alerts as well.

[1] https://github.com/openshift/autoheal/blob/b46f71b591080365a2238362ac2525f697b9fe3a/cmd/autoheal/awx_job.go#L114
[2] https://github.com/openshift/autoheal/blob/b46f71b591080365a2238362ac2525f697b9fe3a/documentation/metrics.md

moolitayer:
Possible solution, post discussion with @zgalor:
- Rename autoheal_actions_initiated_total to autoheal_actions_launched, make it a gauge and add a label status. This gauge will be used from the moment we launch the job and until it is finished(including).
- Expose a counter on requested heals that is incremented when we come to a conclution a heal should take place.
- Expose a metric on received alerts.

moolitayer:
As #59 #70 are merged I think we can close this.
There is one item mentioned here and not implemented which is alert related metrics (e.g count every alert autoheal gets) but I don't think this is a priority right now.

Metrics should be generally usable now.

PR #59, opened 2018-04-22
Add autoheal_actions_requested_total counter

moolitayer:
The new counter will include actions that were rate limited or otherwise failed to run.

Fixes the second item of https://github.com/openshift/autoheal/issues/53

### DIFF
--- a/cmd/autoheal/alerts_worker.go
+++ b/cmd/autoheal/alerts_worker.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"reflect"
 	"regexp"
 
 	"github.com/golang/glog"
@@ -191,6 +192,13 @@ func (h *Healer) runRule(rule *autoheal.HealingRule, alert *alertmanager.Alert) 
 		)
 		return nil
 	}
+
+	// Increment the metric of requested heales.
+	h.actionRequested(
+		reflect.TypeOf(action).Elem().Name(),
+		rule.ObjectMeta.Name,
+		alert.Labels["alertname"],
+	)
 
 	// Process the templates inside the action:
 	template, err := NewObjectTemplateBuilder().
--- a/cmd/autoheal/metrics_exporter.go
+++ b/cmd/autoheal/metrics_exporter.go
@@ -10,6 +10,13 @@ import (
 )
 
 var (
+	actionsRequested = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "autoheal_actions_requested_total",
+			Help: "Number of requested healing actions(including rate limited)",
+		},
+		[]string{"type", "rule", "alert"},
+	)
 	actionsInitiated = prometheus.NewCounterVec(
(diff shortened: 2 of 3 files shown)


cmd/autoheal/metrics_exporter.go
```diff
@@ -39,3 +46,13 @@ func (h *Healer) incrementAwxActions(
 		},
 	).Inc()
 }
+
+func (h *Healer) incrementRequestedActions(type, rule, alert String) {
```
jhernand: I this valid syntax? I think that `type` can't be used as a parameter as it is a reserved word. Also `String` isn't a Go type. Maybe you pushed before committing your latest changes?

cmd/autoheal/alerts_worker.go
```diff
@@ -192,6 +195,13 @@ func (h *Healer) runRule(rule *autoheal.HealingRule, alert *alertmanager.Alert)
 		return nil
 	}
 
+	// Increment the metric of requested heales.
+	h.incrementRequestedActions(
+		actionType,
```
jhernand: Instead of explicitly assigning strings to specify the type you could use `reflect.TypeOf` to find the type and pass it, or just pass type name:

```go
h.incrementRequestedActions(
    reflect.TypeOf(*action).Name(),
    ...,
)
```

That way if in the future we add new types, or modify the names, this wouldn't need to change.

cmd/autoheal/alerts_worker.go
```diff
@@ -179,10 +179,13 @@ func (h *Healer) runRule(rule *autoheal.HealingRule, alert *alertmanager.Alert)
 	// Make a copy of the action so that we can modify it without affecting the rule stored in the
 	// cache:
 	var action interface{}
+	var actionType
```
jhernand: I think this isn't valid syntax, you need to add the type:

```go
var actionType string
```

cmd/autoheal/metrics_exporter.go
nimrodshn: @jhernand if *all* the parameters are of the same type than this is valid.

moolitayer:
Ready for review

cmd/autoheal/alerts_worker.go
```diff
@@ -192,6 +193,13 @@ func (h *Healer) runRule(rule *autoheal.HealingRule, alert *alertmanager.Alert)
 		return nil
 	}
 
+	// Increment the metric of requested heales.
+	h.incrementRequestedActions(
+		reflect.TypeOf(action).Name(),
```
moolitayer: The above currently returns an empty string.

jhernand: That is because in this context the type of `action` is a pointer, and the name of pointer types happens to be empty. Try this:

```go
reflect.TypeOf(action).Elem().Name()
```

moolitayer: Works, thanks!

PR #60, opened 2018-04-23
Update makefile targets doc

yaacov:
**Description**
Currently we do not have "test" and "run" targets in the makefile
We do have "verify-.*" and "build-.*" targets.

This PR update the makefile docs.


yaacov:

PR #61, opened 2018-04-23
Do not try to connect to openshift cluster on default

yaacov:
**Description**

Current "autoheal.yml" example file, try to connect to openshift cluster by default.

In a development enviorment this can be avoided on development system by not sending the optional "credentialsRef" and "tlsRef" vars.

The PR ads an example dev config file that does not require an openshift cluster to run, and add a makefile target to check that the autoheal server can start without need to connect to an openshift cluster.

### DIFF
--- a/hack/run-dev.sh
+++ b/hack/run-dev.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+# This script sets up a go workspace locally and run the autoheal server.
+source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
+
+# Get host plataform
+platform="$(os::build::host_platform)"
+
+# Get platform binary and extention
+binary="${OS_OUTPUT_BINPATH}/${platform}/autoheal"
+if [[ $platform == "windows/amd64" ]]; then
+	binary=("${binary}.exe")
+else
+	binary=("${binary}")
+fi
+
+# Run autoheal server using dev defaults
+"${binary}" server --config-file=examples/autoheal-dev.yml --logtostderr
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@
 #   build-cross: Build the cross compiled release binaries.
 #   build-rpms: Build RPMs only for the Linux AMD64 target.
 #   build-images: Build images from the official RPMs.
+#   run-dev: Run autoheal server using dev defaults.
 
 OUT_DIR = _output
 OS_OUTPUT_GOPATH ?= 1
@@ -126,3 +127,12 @@ build-rpms:
 build-images: build-rpms
 	hack/build-images.sh
 .PHONY: build-images
+
+
+# Run autoheal server using dev defaults.
+#
+# Example:
+#   make run-dev
+run-dev: build
+	hack/run-dev.sh
+.PHONY: run-dev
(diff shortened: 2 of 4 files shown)


yaacov:

zgalor:
I actually think the `autoheal.yml` example should encourage using secrets and default should fail if misconfigured, so I would leave the fields uncommented and if someone wants to use `dev` config they can maintain a local yml file with their own dev config

hack/run-dev.sh
```diff
@@ -0,0 +1,18 @@
...
+# This script sets up a go workspace locally and run the autoheal server.
+source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
+
+# Get host plataform
+platform="$(os::build::host_platform)"
+
+# Get platform binary and extention
+binary="${OS_OUTPUT_BINPATH}/${platform}/autoheal"
+if [[ $platform == "windows/amd64" ]]; then
+	binary=("${binary}.exe")
```
elad661: This is unnecessary - not only we don't really support that platform, it's also unlikely a Windows user will be running bash :)

other than that LGTM

autoheal-dev.yml
```diff
@@ -0,0 +1,119 @@
...
+# limitations under the License.
+#
+
+# This is an example configuration file used for development.
+
+#
+# This section contains the details needed to connect to the AWX or Ansible
+# Tower server that will execute the Ansible playbooks.
+#
+awx:
```
jhernand: Even for development environments you will need at least the `crendentials` section:

```yaml
credentials:
  username: myuser
  password: mypassword
```

Otherwise the connection the AWX server will not work.

autoheal-dev.yml
```diff
@@ -0,0 +1,119 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# This is an example configuration file used for development.
+
+#
+# This section contains the details needed to connect to the AWX or Ansible
```
jhernand: I'd suggest you remove all these duplicated comments from the `autoheal-dev.yml` file, otherwise they will eventually get out of sync. Remove also all the parameters that have default values that you don't need to change.

autoheal-dev.yml
```diff
@@ -0,0 +1,119 @@
...
+  # The URL of the AWX API endpoint, including the `/api` suffix, but not the
+  # `/v1` or `/v2` suffixes.
+  #
+  address: https://127.0.0.1:8080/api
+
+  #
+  # The URL of the proxy server used to connect to the AWX API. If not present
+  # or empty then no proxy server will be used.
+  #
+  proxy: http://127.0.0.1:3128
```
zgalor: this is not required for dev setup

yaacov:
> I actually think the autoheal.yml example should encourage using secrets.

:+1:  left the default autoheal.yml for users example, added a new dev autoheal.yml

autoheal-dev.yml
yaacov: :+1:

autoheal-dev.yml
yaacov: :+1: done

autoheal-dev.yml
yaacov: :+1:  done

yaacov:
@jhernand  I added a develpment section to the Readme file, please re-review.

yaacov:
@nimrodshn @zgalor @moolitayer @elad661 @jhernand . fixed requested fixes, added readme dev section and squashed, Please re-review.

nimrodshn:
Very nice :+1: 
/lgtm

PR #62, opened 2018-04-23
Extract awx client

moolitayer:
Extract the internal awx client with minimal changes.
Extracted client is in:
https://github.com/moolitayer/awx-client-go


jhernand:
Your GitHub repository is perfectly fine, no need to use mine. I'd just suggest to name the package `awxclient` of `go-awx`. That way developers of the Ansible AWX project can clone it without conflicting with the `awx` project.

jhernand:
Also, I think that for that project we should use from the very beginning a semantic versioning schema, and that in the `Gopkg.toml` file of this project we should explicitly require a specific version number.

jhernand:
I also miss the examples in the new repository.

moolitayer:
I choose the name since we are targeting this for inclusion in https://github.com/ansible, including client libraries are named `ansible_tower_client_ruby` & `tower-cli` (python)

The internal awx packge imitates the style of https://github.com/kubernetes/client-go/tree/master/kubernetes

Its not ideal to have `awx-client-go/awx` but it's the best compromise I found

Issue #41, opened 2018-04-11
Automatically reload configuration


jhernand:
Currently the auto-heal service loads the configuration only when it is started. If any changes happen to the configuration file or to the Kubernetes secrets that it uses then the service needs to be restarted. Instead of that the service should watch for changes in the configuration file and in the secrets and reload them automatically. This watching should be implemented internally in the `Config` type. Other parts of the service should have a mechanism to be notified of changes in the configuration, for example:

```go
type ConfigChangeEvent {
}

type ConfigChangeListener func (*ConfigChangeEvent)

config := ...
awxAddress := config.AWX().Address()
config.AddConfigChangeListener(func () {
  awsAddress = config.AWX().Address()
})
```

jhernand:
I assigned this to myself because I can't add @yaacov, but he is the real assignee.

yaacov:
Thanks

yaacov:
closed by #97

PR #64, opened 2018-04-23
Watch config files for changes

yaacov:
**Description**

As a first step for soving https://github.com/openshift/autoheal/issues/41 Watch config files for changes

**Screenshoot**
Editing config files:
config-reload

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -8,6 +8,12 @@
   revision = "3a771d992973f24aa725d07868b467d1ddfceafb"
 
 [[projects]]
+  name = "github.com/fsnotify/fsnotify"
+  packages = ["."]
+  revision = "c2828203cd70a50dcccfb2761f8b1f8ceef9a8e9"
+  version = "v1.4.7"
+
+[[projects]]
   name = "github.com/ghodss/yaml"
   packages = ["."]
   revision = "0ca9ea5df5451ffdf184b4428c902747c2c11cd7"
@@ -169,6 +175,15 @@
   packages = ["."]
   revision = "583c0c0531f06d5278b7d917446061adc344b5cd"
   version = "v1.0.1"
+
+[[projects]]
+  name = "github.com/yaacov/observer"
+  packages = [
+    "observer",
+    "observer/set"
+  ]
+  revision = "e953396b841b524ef5e129d27d826101a92b7dbf"
+  version = "1.4.1"
 
 [[projects]]
   branch = "master"
@@ -431,6 +446,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "7eeddad848f7b0d32edc688635645941c7b8c4721c5c4d2b6f97dc02fe85c5a7"
+  inputs-digest = "695a61275e74a75047dff677310053630c9130742d24280b06c1b3d6bb989e1f"
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/cmd/autoheal/healer.go
+++ b/cmd/autoheal/healer.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/golang/glog"
@@ -34,6 +35,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 
 	"github.com/openshift/autoheal/pkg/alertmanager"
+	"github.com/openshift/autoheal/pkg/apis/autoheal"
 	"github.com/openshift/autoheal/pkg/awxrunner"
 	"github.com/openshift/autoheal/pkg/batchrunner"
 	"github.com/openshift/autoheal/pkg/config"
@@ -55,6 +57,9 @@ type HealerBuilder struct {
 // Prometheus configuration and to start or reload it when there are changes.
 //
 type Healer struct {
+	// The configuration loader.
+	loader *config.Loader
+
 	// The configuration.
 	config *config.Config
 
@@ -74,6 +79,9 @@ type Healer struct {
 
 	// a map of ActionRunner which run awx/batch/etc actions.
 	actionRunners map[ActionRunnerType]ActionRunner
+
+	// Mutex to prevent simultaneous reloads of configuration.
+	reloadMutex *sync.Mutex
 }
 
 // NewHealerBuilder creates a new builder for healers.
@@ -117,13 +125,15 @@ func (b *HealerBuilder) Build() (h *Healer, err error) {
 		err = fmt.Errorf("No configuration file has been provided")
 		return
 	}
-	config, err := config.NewLoader().
+	loader := config.NewLoader().
 		Client(b.k8sClient).
-		Files(b.configFiles).
-		Load()
+		Files(b.configFiles)
+	config, err := loader.Load()
 	if err != nil {
 		return
 	}
+	// Start watching for changes in config files
+	config.Watch()
 
 	// Send to the log a summary of the configuration:
 	glog.Infof("AWX user is '%s'", config.AWX().User())
@@ -140,6 +150,7 @@ func (b *HealerBuilder) Build() (h *Healer, err error) {
 	// Allocate the healer:
 	h = new(Healer)
 	h.k8sClient = b.k8sClient
+	h.loader = loader
 	h.config = config
 	h.actionMemory = actionMemory
 
@@ -152,6 +163,9 @@ func (b *HealerBuilder) Build() (h *Healer, err error) {
 
 	// allocate new action runners
 	h.actionRunners = make(map[ActionRunnerType]ActionRunner)
+
+	// initialize the reload mutex
+	h.reloadMutex = &sync.Mutex{}
 
 	return
 }
@@ -191,6 +205,64 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
 
 	glog.Info("Workers started")
 
+	// Load rules cache from config
+	h.reloadRulesCache()
+
+	// Start the web server:
+	http.Handle("/metrics", metrics.Handler())
+	http.HandleFunc("/alerts", h.handleRequest)
+
+	server := &http.Server{Addr: ":9099"}
+	go server.ListenAndServe()
+	glog.Info("Web server started")
+
+	// Add a listener that will reload the config files
+	// on config file modification.
+	h.config.AddChangeListener(func(e *config.ChangeEvent) {
+		h.onChangeEvent(e)
+	})
+	defer h.config.CloseChangeObserver()
+
+	// Wait till we are requested to stop:
+	<-stopCh
+
+	// Shutdown the web server:
+	err = server.Shutdown(context.TODO())
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// onChangeEvent reload config files and
+// refresh rules in rules cache (by sending "Deleted" + "Added" to queue).
+//
+func (h *Healer) onChangeEvent(e *config.ChangeEvent) {
+	h.reloadMutex.Lock()
+
+	glog.Info("Event: Config file modified")
+	h.loader.Load()
+	h.reloadRulesCache()
+
+	h.reloadMutex.Unlock()
+}
+
+// Reload all rules in rules cache (by sending "Deleted" + "Added" to queue).
+//
+func (h *Healer) reloadRulesCache() {
+	// Send Delete signal to all rules currently in rules cache:
+	h.rulesCache.Range(func(key, value interface{}) bool {
+		rule := value.(*autoheal.HealingRule)
+		change := &RuleChange{
+			Type: watch.Deleted,
+			Rule: rule,
+		}
+		h.rulesQueue.Add(change)
+
+		return true
+	})
+
 	// For each rule inside the configuration create a change and add it to the queue:
 	rules := h.config.Rules()
 	if len(rules) > 0 {
@@ -205,25 +277,6 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
 	} else {
 		glog.Warningf("There are no healing rules in the configuration")
 	}
-
-	// Start the web server:
-	http.Handle("/metrics", metrics.Handler())
-	http.HandleFunc("/alerts", h.handleRequest)
-
-	server := &http.Server{Addr: ":9099"}
-	go server.ListenAndServe()
-	glog.Info("Web server started")
-
-	// Wait till we are requested to stop:
-	<-stopCh
-
-	// Shutdown the web server:
-	err = server.Shutdown(context.TODO())
-	if err != nil {
-		return err
-	}
-
-	return nil
 }
 
 func (h *Healer) handleRequest(response http.ResponseWriter, request *http.Request) {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -22,16 +22,31 @@ import (
 	"bytes"
 	"time"
 
+	"github.com/golang/glog"
+	"github.com/yaacov/observer/observer"
+
 	"github.com/openshift/autoheal/pkg/apis/autoheal"
 )
 
 // Config is a read only view of the configuration of the auto-heal service.
 //
 type Config struct {
-	awx        *AWXConfig
-	throttling *ThrottlingConfig
-	rules      []*autoheal.HealingRule
+	awx                 *AWXConfig
+	throttling          *ThrottlingConfig
+	rules               []*autoheal.HealingRule
+	configFiles         []string // config files loaded by the loader
+	configFilesObserver *observer.Observer
 }
+
+// ChangeEvent contains config change event info
+//
+type ChangeEvent struct {
+	// Empty
+}
+
+// ChangeListener a listener function that can be called when config event change is triggered.
+//
+type ChangeListener func(*ChangeEvent)
 
 // ThrottlingConfig is a read only view of the section of the configuration that describes how to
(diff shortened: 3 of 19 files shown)


yaacov:

nimrodshn:
@yaacov I agree with @jhernand, some changes are due.

yaacov:
> @yaacov I agree with @jhernand, some changes are due.

:+1:  Thanks, adding wip, re-thinking ...

Current thoughts :-)
a. add a config flag to enable the new endpoint (default off)
b. move the clean + update rulles into the `.Load()` function ?

@nimrodshn @jhernand  WDYT ?

jhernand:
I'd suggest the following:

1. The config object remembers the configuration files that it has loaded, and watches them for changes, something similar to what @zgalor did here:

https://github.com/zgalor/prometheus-launcher/blob/bac49db79a5784453a4071716fe8c2611136137c/cmd/launch-n-watch/main.go#L40-L70

2. The config object also watch for changes to the Kubernetes secrets, using the Kubernetes watch API.

3. The config object provides a mechanism to register listeners for configuration change events.

```go
package config

type ChangeEvent ...
type ChangeListener func(*ChangeEvent )

func (c *Config) AddChangeListener(listener ChangeListener) {
  ...
}
```

4. Other parts of the system register for those events, and perform the required actions:

```go
package whatever

config := ...

// Get the initial value of the parameter:
param := config.Param()

// Get notified when the configuration changes:
config.AddChangeListener(func (*event config.Change) {
  // Update the value of the param:
  param = config.Param()

  // Maybe do other actions, for example, clean and populate
  // again the rules cache:
  ...
})
```

No explicit reload support is needed in this approach.

cmd/autoheal/server.go
```diff
@@ -70,6 +71,10 @@ func init() {
 			"directory all the files inside whose names end in .yml or .yaml will be "+
 			"loaded, in alphabetical order.",
 	)
+	serverFlags.BoolVar(
+		&serverWatchConfigFiles, "watch-config-files", false,
+		"Watch config files for changes.",
+	)
```
jhernand: I think this should be always `true` and not configurable.

yaacov: ok :-)

@jhernand @nimrodshn if you are here :-) one Q:

Should we delete all items in `h.rulesCache` each time we reload new version of rules, and re-populate the cache list ?

jhernand: Yes, when the configuration has changed we should remove from the cache all the rules that were loaded from that configuration, and then load them again.

cmd/autoheal/healer.go
```diff
@@ -200,6 +193,44 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
...
+	h.loader.Load()
+
+	// Reload rules qeue
+	h.resetRulesQueue()
+}
+
+// Reset the rules queue, and reload from config.
+//
+func (h *Healer) resetRulesQueue() {
+	// Delete all items in qeue
```
jhernand: The rules worker may need to do something when deleting a rule: closing certain connections to AWX, updating some metrics, etc. So I think it is better if you do this creating a new `RuleChange` for each rule, and letting the rule worker do the actual work.

yaacov: Thanks, will look into it :+1:

yaacov:
@nimrodshn @jhernand  please re-review.

Not sure if the function to reload config does all re-caching / re-freshing needed, please look at [onConfigFilesCanged](https://github.com/openshift/autoheal/pull/64/files#diff-3c6d47d7b38aeac558768f51b2845cadR234) .

> The config object remembers the configuration files that it has loaded, and watches them for changes, something similar to what @zgalor did here:

Done.

> The config object also watch for changes to the Kubernetes secrets, using the Kubernetes watch API.

Not done, I did not find the relevant code :confused: 

> The config object provides a mechanism to register listeners for configuration change events.
> Other parts of the system register for those events, and perform the required actions:

Done using external library: https://github.com/yaacov/observer

cmd/autoheal/healer.go
```diff
@@ -149,6 +158,27 @@ func (b *HealerBuilder) Build() (h *Healer, err error) {
 
 	// initialize active jobs map:
 	h.activeJobs = new(syncmap.Map)
+
```
nimrodshn: @yaacov what about moving this logic from `Healer` to `Loader`? sounds to me like this logic is part of the `Loader`s job an could be cleaner there.
Meaning `Loader` should hold the observer and watch the files and not `Healer`.

yaacov: looking into it :+1:

yaacov: Done :+1:

yaacov:
@nimrodshn  @jhernand . Addressed comments, please review.

Gopkg.lock
```diff
@@ -170,6 +176,15 @@
...
   version = "v1.0.1"
 
+[[projects]]
+  name = "github.com/yaacov/observer"
+  packages = [
+    "observer",
+    "observer/set"
+  ]
+  revision = "e953396b841b524ef5e129d27d826101a92b7dbf"
+  version = "1.4.1"
```
jhernand: 1.4.1 already? This project moves really fast :-) !

pkg/config/loader.go
```diff
@@ -49,15 +50,21 @@ type Loader struct {
 	// The Kubernetes client that will be used to load Kubernetes objects:
 	client kubernetes.Interface
 
-	// The names of the configuration files, in the order that they should be loaded:
+	// The names of the configuration directories and files, in the order that they should be loaded:
 	files []string
 
+	// The names of loaded configuration files:
+	LoadedFiles []string
```
jhernand: Is it necessary to make `LoadedFiles` public?

pkg/config/loader.go
```diff
@@ -49,15 +50,21 @@ type Loader struct {
...
+
 	// The configuration object that is being populated:
 	config *Config
 
 	// The codec that will be used to convert the rules specified in the configuration file into the
 	// types used internally.
 	codec runtime.Codec
+
+	// The config file change watcher.
+	ConfigFilesObserver observer.Observer
```
jhernand: Why does this need to be public?

Not sure how the observer package works, but shouldn't this be a pointer? From what I see in the observer project `Observer` has several fields, and using it directly means that with every call it will it all be copied, which is heavy.

pkg/config/loader.go
```diff
@@ -143,6 +153,13 @@ func (l *Loader) Load() (config *Config, err error) {
 	// Return the created configuration:
 	config = l.config
 
+	// Start/Re-start watching config files for modifications.
+	err = l.ConfigFilesObserver.Watch(l.LoadedFiles)
+	if err != nil {
+		glog.Errorf("Error watching config files: %v.", err)
```
jhernand: In the rest of the code of the project we log messages don't end with punctuation. Can you do the same?

pkg/config/loader.go
```diff
@@ -72,6 +79,9 @@ func NewLoader() *Loader {
 	v1alpha2.AddToScheme(scheme)
 	l.codec = serializer.NewCodecFactory(scheme).LegacyCodec()
 
+	// Start a change watcher over loaded config files.
+	l.ConfigFilesObserver.Open()
```
jhernand: When is this closed?

cmd/autoheal/healer.go
```diff
@@ -188,6 +182,16 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
 	go server.ListenAndServe()
 	glog.Info("Web server started")
 
+	// Add a listener that will reload the config files
+	// on config file modification.
+	h.loader.ConfigFilesObserver.AddListener(func(e interface{}) {
```
jhernand: This doesn't feel right: the healer should not know that the loader has a `ConfigFilesObserver` field as that is an implementation detail. The `AddListener` method (or similiar) should be a method of the config, not of the loader, and it should't leak the details of the underlying implementation.

cmd/autoheal/healer.go
```diff
@@ -200,6 +204,47 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
 	return nil
 }
 
+// Reload all rules in rules cache (by sending "Deleted" + "Added" to queue).
+//
+func (h *Healer) onConfigFilesCanged() {
```
jhernand: I think that using the word _files_ here indicates that this is two low level. Here the healer should not care about the source of the configuration: it may be files, it may be something else. This method should be named `onConfigChanged` and the configuration should have already been reloaded at this point.

pkg/config/loader.go
yaacov: thanks ! it was when the logic was in healer ... now it's not :+1:

pkg/config/loader.go
yaacov: looking into it

yaacov:
1. `loadedFiles` and `configFilesObserver`   are now local.
2. log messages don't end with punctuation
3. add defer close observer
4. `onConfigFilesCanged` is now `onConfigChanged`

@jhernand ready for ( next :-) ) review

pkg/config/loader.go
```diff
@@ -43,21 +44,36 @@ import (
...
+
 	// The configuration object that is being populated:
 	config *Config
 
 	// The codec that will be used to convert the rules specified in the configuration file into the
 	// types used internally.
 	codec runtime.Codec
+
+	// The config file change watcher.
+	configFilesObserver *observer.Observer
```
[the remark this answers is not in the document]
nimrodshn: @jhernand actually I initially thought it should go into the `Loader` struct - and `Config` struct is only a container for information.. That's why I told @yaacov to move that logic here.
It's clear to me it shouldn't be in `Healer`..

yaacov:
a. Moved the change watcher to config.
b. Loader knows what files to watch, but config does the actual watching.
c. Healer has a listener function that get called by config when change event is triggered.

@nimrodshn @jhernand next review please :-)

cmd/autoheal/healer.go
```diff
@@ -200,6 +208,52 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
 	return nil
 }
 
+// onChangeEvent reload config files and
+// refresh rules in rules cache (by sending "Deleted" + "Added" to queue).
+//
+func (h *Healer) onChangeEvent(e *config.ChangeEvent) {
+	mutex := &sync.Mutex{}
```
jhernand: If you create the mutex inside the function then it has no effect, as each goroutine will use its own mutex. If you really need to make sure that two events aren't processed in parallel then you will need to do something like this:

```go
type Healer {
  ...
  // Mutex to prevent simultaneous reloads of configuration.
  reloadMutex *sync.Mutex
  ...
}

func (h *Healer) onChangeEvent(e *config.ChangeEvent) {
  h.reloadMutex.Lock()
  defer h.reloadMutex.Unlock()

  ...
}
```

cmd/autoheal/healer.go
```diff
@@ -200,6 +208,52 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
...
+
+	mutex.Lock()
+
+	glog.Infof("Received a config modification event: %s", e.Info)
+	h.loader.Load()
```
jhernand: I still think that when this function is called the configuration should have been reloaded already. This method should just need what is needed to apply that new configuration, which in this case is only reloading the rules, if I understand correctly.

pkg/config/config.go
```diff
@@ -22,17 +22,31 @@ import (
...
+	awx                 *AWXConfig
+	throttling          *ThrottlingConfig
+	rules               []*autoheal.HealingRule
+	configFilesObserver *observer.Observer
 }
 
+// ChangeEvent contains config change event info
+//
+type ChangeEvent struct {
+	Info string
```
jhernand: Can we make `Info` private and, if we want to print it, add a `String` method?

```go
type ChangeEvent struct {
  info string
}

func (e *ChangeEvent) String() {
  return e.info
}
```

That way the user of the `ChangeEvent` type doesn't need to know (and depend on) the fact that it has an `Info` field.

pkg/config/loader.go
```diff
@@ -49,9 +49,12 @@ type Loader struct {
 	// The Kubernetes client that will be used to load Kubernetes objects:
 	client kubernetes.Interface
 
-	// The names of the configuration files, in the order that they should be loaded:
+	// The names of the configuration directories and files, in the order that they should be loaded:
 	files []string
 
+	// The names of loaded configuration files:
+	loadedFiles []string
+
```
jhernand: Can we move this field to the config type?

cmd/autoheal/healer.go
```diff
@@ -188,6 +183,19 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
...
+	defer h.config.CloseChangeObserver()
+
+	// Add a listener that will reload the config files
+	// on config file modification.
+	h.config.AddChangeListener(func(e *config.ChangeEvent) {
+		h.onChangeEvent(e)
+	})
+
+	// Start watching loaded config files for changes
+	h.loader.Watch()
```
jhernand: This should also be hidden inside the implementation of the config type.

pkg/config/loader.go
yaacov: :+1: woo, that will solve the watch problem too, on it

yaacov:
Last iteration ? :-)
@nimrodshn @jhernand please review.

cmd/autoheal/healer.go
yaacov: I do not know how to trigger the listeners in order, in a nice way  ...

My problem:
On event we need:
1. loader.load
2. reload rules cache

Solutions I thought about:
a. we can register two listeners: but we do not have a mechanism to ensure one will run after the other.
b. we can add a new observer to the loader, that will emit "reload rules cache" once loader.load is finished: but I think that's over engineering it ...

cmd/autoheal/healer.go
```diff
@@ -188,6 +189,19 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
 	go server.ListenAndServe()
 	glog.Info("Web server started")
 
+	// Start a change watcher over loaded config files.
+	h.config.OpenChangeObserver()
+	defer h.config.CloseChangeObserver()
```
jhernand: I still think that this should be hidden in the implementation of the config. Why do we need to do it explicitly here?

cmd/autoheal/healer.go
```diff
@@ -188,6 +189,19 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
...
+
+	// Add a listener that will reload the config files
+	// on config file modification.
+	h.config.AddChangeListener(func(e *config.ChangeEvent) {
+		h.onChangeEvent(e)
+	})
+
+	// Start watching loaded config files for changes
+	h.config.Watch()
```
jhernand: Same, this should be hidden in the implementation of the config.

pkg/config/config.go
```diff
@@ -20,19 +20,47 @@ package config
...
+
+// ChangeEvent contains config change event info
+//
+type ChangeEvent struct {
+	info string
+}
+
+// Info set the event info
+//
+func (e *ChangeEvent) Info(info string) ChangeEvent {
```
jhernand: Why do we need this public method? What outside of this package is calling it?

pkg/config/loader.go
```diff
@@ -213,6 +213,9 @@ func (l *Loader) mergeFile(file string) error {
 		}
 	}
 
+	// Add the file to loaded config files list:
+	l.config.AppendFile(file)
```
jhernand: Here we can safely manipulate the private fields of the config, I think we don't need this `AppendFile` method, specially not if it is public.

jhernand:
@yaacov last iteration minus delta, where delta tends to zero.

yaacov:
bad rebase ... fixing

pkg/config/config.go
```diff
@@ -141,3 +156,41 @@ func (t *ThrottlingConfig) Interval() time.Duration {
 func (c *Config) Rules() []*autoheal.HealingRule {
 	return c.rules
 }
+
+// OpenChangeObserver open the change obeserver channels
+//
+func (c *Config) OpenChangeObserver() {
```
jhernand: This can now be private, or can be inlined directly in the place where it is called.

pkg/config/config.go
```diff
@@ -141,3 +156,41 @@ func (t *ThrottlingConfig) Interval() time.Duration {
...
+	c.configFilesObserver = new(observer.Observer)
+	c.configFilesObserver.Open()
+}
+
+// CloseChangeObserver close the change obeserver channels
+//
+func (c *Config) CloseChangeObserver() {
```
jhernand: I think that now this is never called. Do we still need it?

yaacov: I had a bad rebase ... it is called in the healer `defer` to program exit

fixing the rebase now ... lots of code missing in healer ...

pkg/config/config.go
yaacov: yes, done ... but fixing rebase ...

yaacov:
@jhernand , had a bad rebase, fixed now :tada: 

@jhernand @nimrodshn please review.

jhernand:
@yaacov the change in itself looks good, and I am inclined to merge it, but we need to make sure that the dependencies are updated correctly. To do so we need to include in this pull request a commit to update the `vendor` directory, the result of running `dep ensure`. That should add to the `vendor` directory the code of the `observer` package, at least.

nimrodshn:
@yaacov very nice. 
/lgtm
/approve

PR #65, opened 2018-04-23
Proposal: Extract ActionRunner interface

nimrodshn:
## What this does?

- Extracting run*Job to a ~~JobRunner~~ ActionRunner interface and adds structs to implement this interface.
- Adds a ~~JobRunner~~ ActionRunner map to healer.

## Why do this?

1. This extracts away methods and fields from healer - an already very heavy struct and creates ~~JobRunners~~ ActionRunner specific method and implementations which healer shouldn't be aware of.
2. This also, allows to extend ~~JobRunners~~ ActionRunner to any number and variety (If someday we would like to add more options more targets for jobs)
3. And most importantly; it allows me to stub for `RunJob` and by that write tests for `alerts_worker` and other parts of healer which currently not testable :cry: .

cc: @yaacov @moolitayer @jhernand

### DIFF
--- a/cmd/autoheal/alerts_worker.go
+++ b/cmd/autoheal/alerts_worker.go
@@ -22,11 +22,11 @@ import (
 	"regexp"
 
 	"github.com/golang/glog"
-	batch "k8s.io/api/batch/v1"
-	"k8s.io/apimachinery/pkg/util/runtime"
-
 	"github.com/openshift/autoheal/pkg/alertmanager"
 	"github.com/openshift/autoheal/pkg/apis/autoheal"
+	"github.com/openshift/autoheal/pkg/metrics"
+	batch "k8s.io/api/batch/v1"
+	"k8s.io/apimachinery/pkg/util/runtime"
 )
 
 func (h *Healer) runAlertsWorker() {
@@ -194,7 +194,7 @@ func (h *Healer) runRule(rule *autoheal.HealingRule, alert *alertmanager.Alert) 
 	}
 
 	// Increment the metric of requested heales.
-	h.actionRequested(
+	metrics.ActionRequested(
 		reflect.TypeOf(action).Elem().Name(),
 		rule.ObjectMeta.Name,
 		alert.Labels["alertname"],
@@ -227,9 +227,9 @@ func (h *Healer) runRule(rule *autoheal.HealingRule, alert *alertmanager.Alert) 
 	// Execute the action:
 	switch typed := action.(type) {
 	case *autoheal.AWXJobAction:
-		err = h.runAWXJob(rule, typed, alert)
+		err = h.actionRunners[ActionRunnerTypeAWX].RunAction(rule, typed, alert)
 	case *batch.Job:
-		err = h.runBatchJob(rule, typed, alert)
+		err = h.actionRunners[ActionRunnerTypeBatch].RunAction(rule, typed, alert)
 	default:
 		err = fmt.Errorf(
 			"Don't know how to execute action of type '%T'",
--- a/cmd/autoheal/healer.go
+++ b/cmd/autoheal/healer.go
@@ -34,8 +34,11 @@ import (
 	"k8s.io/client-go/util/workqueue"
 
 	"github.com/openshift/autoheal/pkg/alertmanager"
+	"github.com/openshift/autoheal/pkg/awxrunner"
+	"github.com/openshift/autoheal/pkg/batchrunner"
 	"github.com/openshift/autoheal/pkg/config"
 	"github.com/openshift/autoheal/pkg/memory"
+	"github.com/openshift/autoheal/pkg/metrics"
 )
 
 // HealerBuilder is used to create new healers.
@@ -69,8 +72,8 @@ type Healer struct {
 	// Executed actions will be stored here in order to prevent repeated execution.
 	actionMemory *memory.ShortTermMemory
 
-	// The AWX active jobs
-	activeJobs *syncmap.Map
+	// a map of ActionRunner which run awx/batch/etc actions.
+	actionRunners map[ActionRunnerType]ActionRunner
 }
 
 // NewHealerBuilder creates a new builder for healers.
@@ -147,8 +150,9 @@ func (b *HealerBuilder) Build() (h *Healer, err error) {
 	h.rulesQueue = workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "rules")
 	h.alertsQueue = workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "alerts")
 
-	// initialize active jobs map:
-	h.activeJobs = new(syncmap.Map)
+	// allocate new action runners
+	h.actionRunners = make(map[ActionRunnerType]ActionRunner)
+
 	return
 }
 
@@ -162,7 +166,29 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
 	// Start the workers:
 	go wait.Until(h.runRulesWorker, time.Second, stopCh)
 	go wait.Until(h.runAlertsWorker, time.Second, stopCh)
-	go wait.Until(h.runActiveJobsWorker, h.config.AWX().JobStatusCheckInterval(), stopCh)
+
+	// Start action runners
+	awxRunner, err := awxrunner.NewBuilder().
+		Config(h.config.AWX()).
+		StopCh(stopCh).
+		Build()
+
+	if err != nil {
+		glog.Warningf("Error building AWX Runner: %s", err)
+	}
+
+	batchRunner, err := batchrunner.NewBuilder().
+		KuberenetesClient(h.k8sClient).
+		Build()
+
+	if err != nil {
+		glog.Warningf("Error building Batch Runner: %s", err)
+	}
+
+	// initiailize runners.
+	h.actionRunners[ActionRunnerTypeAWX] = awxRunner
+	h.actionRunners[ActionRunnerTypeBatch] = batchRunner
+
 	glog.Info("Workers started")
 
 	// For each rule inside the configuration create a change and add it to the queue:
@@ -181,7 +207,7 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
 	}
 
 	// Start the web server:
-	http.Handle("/metrics", h.metricsHandler())
+	http.Handle("/metrics", metrics.Handler())
 	http.HandleFunc("/alerts", h.handleRequest)
 
 	server := &http.Server{Addr: ":9099"}
@@ -192,7 +218,7 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
 	<-stopCh
 
 	// Shutdown the web server:
-	err := server.Shutdown(context.TODO())
+	err = server.Shutdown(context.TODO())
 	if err != nil {
 		return err
 	}
--- a/pkg/awxrunner/awx_action_runner.go
+++ b/pkg/awxrunner/awx_action_runner.go
@@ -14,34 +14,73 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package awxrunner
 
 import (
 	"fmt"
 
 	"github.com/golang/glog"
+	"golang.org/x/sync/syncmap"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/moolitayer/awx-client-go/awx"
 	"github.com/openshift/autoheal/pkg/alertmanager"
 	"github.com/openshift/autoheal/pkg/apis/autoheal"
+	"github.com/openshift/autoheal/pkg/config"
+	"github.com/openshift/autoheal/pkg/metrics"
 )
 
-func (h *Healer) runAWXJob(rule *autoheal.HealingRule, action *autoheal.AWXJobAction, alert *alertmanager.Alert) error {
-	var err error
+type Builder struct {
+	config *config.AWXConfig
 
+	stopCh <-chan struct{}
+}
+
+type Runner struct {
+	config *config.AWXConfig
+
+	activeJobs *syncmap.Map
+}
+
+func NewBuilder() *Builder {
+	return new(Builder)
+}
+
+func (b *Builder) Config(config *config.AWXConfig) *Builder {
+	b.config = config
+	return b
+}
+
+func (b *Builder) StopCh(stopCh <-chan struct{}) *Builder {
+	b.stopCh = stopCh
+	return b
+}
+
+func (b *Builder) Build() (*Runner, error) {
+	runner := &Runner{
+		config:     b.config,
+		activeJobs: new(syncmap.Map),
+	}
+	go wait.Until(runner.runActiveJobsWorker, runner.config.JobStatusCheckInterval(), b.stopCh)
+	return runner, nil
+}
+
+func (r *Runner) RunAction(rule *autoheal.HealingRule, action interface{}, alert *alertmanager.Alert) error {
+	var err error
+	awxAction := action.(*autoheal.AWXJobAction)
 	// Get the AWX connection details from the configuration:
-	awxAddress := h.config.AWX().Address()
-	awxProxy := h.config.AWX().Proxy()
-	awxUser := h.config.AWX().User()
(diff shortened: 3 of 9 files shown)


cmd/autoheal/alerts_worker.go
```diff
@@ -219,9 +219,10 @@ func (h *Healer) runRule(rule *autoheal.HealingRule, alert *alertmanager.Alert)
 	// Execute the action:
 	switch typed := action.(type) {
 	case *autoheal.AWXJobAction:
```
yaacov: can we have the rule have a "runner" field, so we can just do `rule.run()` ?
I think it will be much more fun then this case ...

nimrodshn: Sure, good suggestion!

cmd/autoheal/active_jobs_worker.go
```diff
@@ -21,15 +21,15 @@ import (
 	"k8s.io/apimachinery/pkg/util/runtime"
 )
 
-func (h *Healer) runActiveJobsWorker() {
+func (r AWXJobRunner) runActiveJobsWorker() {
```
jhernand: `r` should be a pointer, otherwise you will be copying the contents of the struct each time you call this method (and any other method that has the struct as receiver):

```go
func (r *AWXJobRunner) runActiveJobsWorker() {
  ...
}
```

cmd/autoheal/alerts_worker.go
jhernand: I'd prefer to avoid that and keep the rule object as a pure container of data, without logic.

cmd/autoheal/alerts_worker.go
```diff
@@ -219,9 +219,10 @@ func (h *Healer) runRule(rule *autoheal.HealingRule, alert *alertmanager.Alert)
...
-		err = h.runAWXJob(rule, typed, alert)
+		err = h.jobRunners[AWX].RunJob(rule, typed, alert)
+
 	case *batch.Job:
-		err = h.runBatchJob(rule, typed, alert)
+		err = h.jobRunners[BATCH].RunJob(rule, typed, alert)
```
jhernand: What is the benefit of this map if we always index it with constants? Wouldn't it be simpler to have two fields?

```go
type Healer struct {
  ...
  JobRunner awxRunner;
  JobRunner batchRunner;
  ...
}

File: cmd/autoheal/awx_job.go [hunk 4]
```diff
@@ -162,3 +171,7 @@ func (h *Healer) checkAWXJobStatus(jobID int) (finished bool, err error) {
 
 	return
 }
+
+func (r AWXJobRunner) AWX() *config.AWXConfig {
```
jhernand: Using this method makes calls shorter, but also less clear: it is no longer explicit that you are using the configuration. I'd suggest to avoid that.

cmd/autoheal/healer.go
```diff
@@ -147,8 +147,17 @@ func (b *HealerBuilder) Build() (h *Healer, err error) {
 	h.rulesQueue = workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "rules")
 	h.alertsQueue = workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "alerts")
 
-	// initialize active jobs map:
-	h.activeJobs = new(syncmap.Map)
+	h.jobRunners = map[JobRunnerType]JobRunner{
```
jhernand: If we are using an interface to hide the implementation details of each job runner, then we should also hide the implementation details of the creation of the job runner, we should not access directly internal fields like `config` or `activeJobs`. To enforce this I'd suggest that you move the implementations of the job runners to different packages, and make everything private, unless really needed. That will force you to create a clean interface between the healer and the job runner.

cmd/autoheal/job_runner.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	alertmanager "github.com/openshift/autoheal/pkg/alertmanager"
+	autoheal "github.com/openshift/autoheal/pkg/apis/autoheal/v1alpha2"
+)
+
+type JobRunnerType int
+
+const (
+	AWX   JobRunnerType = 0
```
jhernand: I'd prefer to avoid this kind of constants, as the fact that there are different types of job runners is already indicated by the existence of the corresponding Go types. If the constants are needed then please consider using `reflect.TypeOf` or `reflect.TypeOf(...).Name()` instead.

Also, for constants try to use camel case, and add a prefix that corresponds to the type name:

```go
const (
   JobRunnerTypeAWX JobRunnerType = iota
   JobRunnerTypeBatch
)
```

cmd/autoheal/job_runner.go
```diff
@@ -0,0 +1,33 @@
...
+	BATCH JobRunnerType = 1
+)
+
+type JobRunner interface {
```
jhernand: I am not 100% sure if `JobRunner` is the right name for this interface. Currently it is running either an AWX job or a Kubernetes batch Job, thus the _job_ prefix makes sense. But in general what it runs is a _healing action_, which may be other thing. For example, if in the future we add support for Ansible Runner then it will be running an Ansible Playbook, not a _job_. So consider renaming it to `HealingActionRunner`, or something similar.

cmd/autoheal/alerts_worker.go
nimrodshn: I just thought it would be easier to scale with a map if we eventually in the future add Runners..

cmd/autoheal/healer.go
nimrodshn: Working on it :+1:

cmd/autoheal/job_runner.go
nimrodshn: Sounds good :+1:

cmd/autoheal/alerts_worker.go
nimrodshn: I tend to agree with @jhernand

cmd/autoheal/healer.go
```diff
@@ -147,8 +147,17 @@ func (b *HealerBuilder) Build() (h *Healer, err error) {
...
+		AWX: AWXJobRunner{
+			config: h.config.AWX(),
```
moolitayer: When constructing the implementations It would be clean if they each get the entire configuration, and select the pieces relevant to them.

cmd/autoheal/healer.go
```diff
@@ -147,8 +147,17 @@ func (b *HealerBuilder) Build() (h *Healer, err error) {
...
+			// initialize active jobs map:
+			activeJobs: new(syncmap.Map),
```
moolitayer: This does not have to be part this PR, but it would be cleaner if the creator of an AWXJobRunner would not have to supply this map. It could be done in a constructor.

cmd/autoheal/job_runner.go
moolitayer: @nimrodshn I gave the same comment when we talked about this but did't explain it as well...

cmd/autoheal/healer.go
nimrodshn: Yeah thats a good comment I just noticed that going over the code again :+1:

cmd/autoheal/healer.go
nimrodshn: Nice thanks @moolitayer

nimrodshn:
@moolitayer @jhernand @yaacov  Addressing you're comments I have done the following:

1. Extracted the `JobRunners` to `pkg/job_runners`
2. Extracted `metrics_exporter.go` from healer to a separate `pkg` so the runners can use it.
3. Fixed existing tests to use this `Runner` pattern.
4. Tested that healer works well with this new pattern.

Things to do in following PR's moving forward:
1. Move `Init` method *outside* the structs implementing `Runner` and into the `Runner` itself thus having a clean interface.
2. Other ideas are welcome here to enhance this patters :grinning: .

yaacov:
LGTM :+1:

nimrodshn:
@jhernand @moolitayer mind taking a look?

pkg/job_runner/awx_job_runner/awx_job_runner.go
```diff
@@ -14,34 +14,57 @@ See the License for the specific language governing permissions and
...
+
+	activeJobs *syncmap.Map
+}
+
+func NewAWXJobRunner() *AWXJobRunner {
+	awxJubRunner := new(AWXJobRunner)
+	return awxJubRunner
+}
 
+func (r *AWXJobRunner) Init(config *config.Config, stopCh <-chan struct{}) *AWXJobRunner {
```
jhernand: This kind of constructor in two steps is problematic, in my opinion, specially the fact that the `Init` method receives multiple parameters. What if we need or want to add a new optional parameter in the future? We will have to change all the points where this method is called. That isn't very relevant in this case, because it is used only internally, but in general I prefer to use a builder:

```go
type Builder {
  ...
}

func NewBuilder() *Builder {
  return new(Builder)
}

// Config sets the configuration that will be used by the
// runner.
func (b *Builder) Config(config *config.Config) *Builder {
  b.config = config
  return b
}

// StopChannel sets the channel where the runner will wait
// for the signal to stop its background processing.
func (b *Builder) StopChannel(ch <-chan struct{}) *Builder {
  b.stopCh = ch
  return b
}

// Build creates a new runner using the details stored in
// the builder.
func (b *Builder) Build() (runner *AWXJobRunner, err error) {
  // Create and populate the runner, and also check for
  // possible errors, like missing mandatory parameters,
  // etc.
  runner = ...

  return
}
```

This way the user of this package can create the runner like this:

```go
runner, err := awxrunner.NewBuilder().
  Config(config).
  StopChannel(stopCh).
  Build()
if err != nil {
  ...
}
```

If we eventually decide to add optional parameters to that constructor, that code doesn't need to change.

Can you use this mechanism for both kinds of runners?

pkg/job_runner/awx_job_runner/awx_job_runner.go
```diff
@@ -14,34 +14,57 @@ See the License for the specific language governing permissions and
...
+	var err error
+	awxAction := action.(*autoheal.AWXJobAction)
 	// Get the AWX connection details from the configuration:
-	awxAddress := h.config.AWX().Address()
-	awxProxy := h.config.AWX().Proxy()
-	awxUser := h.config.AWX().User()
-	awxPassword := h.config.AWX().Password()
-	awxCA := h.config.AWX().CA()
-	awxInsecure := h.config.AWX().Insecure()
+	awxAddress := r.AWX().Address()
```
jhernand: I think that this `AWX` method unnecessary hides the fact that we are accessing the configuration. Please avoid it.

pkg/job_runner/awx_job_runner/awx_job_runner.go
```diff
@@ -167,3 +190,7 @@ func (h *Healer) checkAWXJobStatus(jobID int) (finished bool, err error) {
 
 	return
 }
+
+func (r *AWXJobRunner) AWX() *config.AWXConfig {
```
jhernand: I think this method isn't necessary.

pkg/job_runner/awx_job_runner/awx_job_runner.go
nimrodshn: Yeah very nice! Actoually I never really agreed with this approach in the code, but now i'm sold :+1: :trophy:

cmd/autoheal/healer.go
```diff
@@ -162,7 +166,16 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
 	// Start the workers:
 	go wait.Until(h.runRulesWorker, time.Second, stopCh)
 	go wait.Until(h.runAlertsWorker, time.Second, stopCh)
-	go wait.Until(h.runActiveJobsWorker, h.config.AWX().JobStatusCheckInterval(), stopCh)
+
+	// Start job runners
+	h.actionRunners[ActionRunnerTypeAWX] = awxrunner.NewAWXActionRunnerBuilder().
+		Config(h.config.AWX()).
+		StopCh(stopCh).
+		Build()
```
jhernand: I think we should check the errors returned by the `Build` methods.

pkg/awxrunner/awx_action_runner.go
```diff
@@ -14,34 +14,73 @@ See the License for the specific language governing permissions and
...
 	"github.com/moolitayer/awx-client-go/awx"
 	"github.com/openshift/autoheal/pkg/alertmanager"
 	"github.com/openshift/autoheal/pkg/apis/autoheal"
+	"github.com/openshift/autoheal/pkg/config"
+	"github.com/openshift/autoheal/pkg/metrics"
 )
 
-func (h *Healer) runAWXJob(rule *autoheal.HealingRule, action *autoheal.AWXJobAction, alert *alertmanager.Alert) error {
-	var err error
+type AWXActionRunnerBuilder struct {
```
jhernand: As this has its own package now I think you can rename it to just `Builder`. Same for the runner, it can be just `Runner`.

pkg/awxrunner/awx_action_runner.go
```diff
@@ -167,3 +206,7 @@ func (h *Healer) checkAWXJobStatus(jobID int) (finished bool, err error) {
 
 	return
 }
+
+func (r *AWXActionRunner) AWX() *config.AWXConfig {
```
jhernand: I still think that this method isn't needed, and using it hides the fact that we are using the configuration. Can you remove this method and use directly `config...`?

pkg/batchrunner/batch_action_runner.go
```diff
@@ -14,28 +14,60 @@ See the License for the specific language governing permissions and
...
+	b.k8sClient = k8sClient
+	return b
+}
+
+func (b *BatchActionRunnerBuilder) StopCh(stopCh <-chan struct{}) *BatchActionRunnerBuilder {
+	b.stopCh = stopCh
+	return b
+}
+
+func (b *BatchActionRunnerBuilder) Build() *BatchActionRunner {
```
jhernand: The build method should return the built object an and an error. We may not need the error today, but it will most likely needed in the future. Please add it.

pkg/awxrunner/awx_action_runner.go
```diff
@@ -14,34 +14,73 @@ See the License for the specific language governing permissions and
...
+	b.config = config
+	return b
+}
+
+func (b *AWXActionRunnerBuilder) StopCh(stopCh <-chan struct{}) *AWXActionRunnerBuilder {
+	b.stopCh = stopCh
+	return b
+}
 
+func (b *AWXActionRunnerBuilder) Build() *AWXActionRunner {
```
jhernand: Build should return the result and an error.

pkg/metrics/metrics_exporter.go
```diff
@@ -24,15 +24,19 @@ var (
 	)
 )
 
-func (h *Healer) metricsHandler() http.Handler {
```
jhernand: As this has its own package now, you can simplify the names, just `Handler`, for example. That way the user of this package will write `metrics.Handler` which reads nicely.

pkg/awxrunner/awx_action_runner.go
nimrodshn: Yeah I forgot to remove it :cry:

pkg/metrics/metrics_exporter.go
nimrodshn: nice :+1:

nimrodshn:
@jhernand Please take a look :+1:

cmd/autoheal/healer.go
```diff
@@ -162,7 +166,24 @@ func (h *Healer) Run(stopCh <-chan struct{}) error {
 	// Start the workers:
 	go wait.Until(h.runRulesWorker, time.Second, stopCh)
 	go wait.Until(h.runAlertsWorker, time.Second, stopCh)
-	go wait.Until(h.runActiveJobsWorker, h.config.AWX().JobStatusCheckInterval(), stopCh)
+
+	// Start job runners
+	awxRunner, err := awxrunner.NewBuilder().
+		Config(h.config.AWX()).
+		StopCh(stopCh).
+		Build()
```
jhernand: We need to check `err` here as well, otherwise if this fails but the next does't then we don't detect it.

nimrodshn:
@jhernand Done.

Issue #39, opened 2018-04-11
Send all labels and annotations as `extraVars` by default


jhernand:
Currently the auto heal service can send to the AWX job a set of `extraVars` that are defined as a JSON document:

```yaml
awxJob:
  template: "My template"
  extraVars: |-
    {
      "myvar": "myvalue",
      "yourvar": "yourvalue",
    }
```

This is very useful to send values or labels or annotations of the alert, for example, to send the value of the `instance` label:

```yaml
awxJob:
  template: "My template"
  extraVars: |-
    {
      "instance": "{{ $labels.instance }}",
    }
```

It is so useful that it should be the default: if the `extraVars` field isn't used then we should automatically populate it with all the labels and annotations of the alert. For example, if the alert is like this:

```yaml
labels:
  instance: 192.168.100.7:9100
  job: node-exporter-123
annotations:
  message: "Node '192.168.100.7:9100' is down"
```

Then we should automatically populate the `extraVars` field like this:

```yaml
extraVars: |-
  {
    "labels": {
      "instance": "192.168.100.7:9100",
      "job": "node-exporter-123"
    },
    "annotations": {
      "message": "Node '192.168.100.7
    }
  }
```

Actually we should probably just provide the full alert description:

```yaml
extraVars: |-
  {
    "alerts": [
     ...
    ]
  }
```

This should be compatible with other custom `extraVars` that the user may want to add. For example, the following action:

```yaml
awxJob:
  template: "My template"
  extraVars:
    myvar: myvalue
    yourvalue: yourvalue
```

Should be equivalent to this:

```yaml
awxJob:
  template: "My template"
  extraVars: |-
    {
      "myvar": "myvalue",
      "yourvar": "yourvalue",
      "alerts": [...]
    }
```

jhernand:
Another thing, related to this, is that in order to send the `extraVars` via the API the job template has to be configured with `Prompt on launch` flag set for the extra variables, otherwise whatever the client sends is ignored. We should probably check the value of this flag before executing the job, and either fail the execution completely or, at least, send a warning to the log explaining that the template isn't configured correctly and that it will ignore the values.

jhernand:
@zgalor I just noticed that currently we aren't managing correctly the `extraVars` field. The AWX API expects this to be a JSON document. For example, if you want to send the `my` and `your` variables then what you need to POST to launch a job is something like this:

```json
{
  "extraVars": {
    "my": "...",
    "your": "...",
  }
}
```

[...]

PR #67, opened 2018-04-24
Fix the way extraVars are sent to AWX.

zgalor:
Change extraVars field from string to JSON, as AWX API expects it.
In autoheal.yml change the field from YAML string to JSON.
Add map support in object_template.go processValue func.
This changed required changing the processString function not to change the string inplace but to return it,
Also the recursion in processValue had to return and set new values back to the map, since reflect API doesn't allow changing in place.

This fixes https://github.com/openshift/autoheal/issues/39#issuecomment-380708563


cmd/autoheal/object_template.go
```diff
@@ -121,33 +121,55 @@ func (t *ObjectTemplate) Process(object interface{}, data interface{}) error {
...
+	}
+
+	return err
 }
 
-func (t *ObjectTemplate) processValue(value reflect.Value, data interface{}) error {
-	var err error
-	if value.IsValid() {
-		switch value.Kind() {
+func (t *ObjectTemplate) processValue(inputValue reflect.Value, data interface{}) (val reflect.Value, err error) {
```
zgalor: I am not so happy with what I did here.
However it works, and I couldn't find a better way to do it.

Would appreciate your reviews 
@jhernand @moolitayer @yaacov @nimrodshn

cmd/autoheal/object_template.go
```diff
@@ -121,33 +121,55 @@ func (t *ObjectTemplate) Process(object interface{}, data interface{}) error {
...
+				if err == nil {
+					// update the processed vaule in the map
+					val.SetMapIndex(k, v)
+				}
+			}
 		case reflect.Struct:
-			for i, n := 0, value.NumField(); i < n && err == nil; i++ {
-				err = t.processValue(value.Field(i), data)
+			for i, n := 0, val.NumField(); i < n && err == nil; i++ {
+				_, err = t.processValue(val.Field(i), data)
```
jhernand: We also need to explicitly check for `err != nil` here. That was incorrect before this pull request already.

vendor/github.com/moolitayer/awx-client-go/awx/job_template_launch_resource.go
```diff
@@ -20,6 +20,8 @@ limitations under the License.
 package awx
```
jhernand: Note that the AWX client has bee moved to a different repository: https://github.com/moolitayer/awx-client-go. We will need to do these changes there, make a new release, and then update this project to use that new release.

vendor/github.com/moolitayer/awx-client-go/awx/job_template_launch_resource.go
```diff
@@ -76,18 +78,24 @@ func (r *JobTemplateLaunchGetResponse) JobTemplateData() *JobTemplate {
 type JobTemplateLaunchPostRequest struct {
 	Request
 
-	extraVars string
+	extraVars map[string]interface{}
 }
 
-func (r *JobTemplateLaunchPostRequest) ExtraVars(value string) *JobTemplateLaunchPostRequest {
+func (r *JobTemplateLaunchPostRequest) ExtraVars(value map[string]interface{}) *JobTemplateLaunchPostRequest {
```
jhernand: I think that in addition to this `ExtraVars` method we should also have another `ExtraVar` method that receives only the variable name and the value, so that users of the AWX client can do something like this:

```go
response := resource.NewJobTemplateLaunch().
  ExtraVar("myvar", "myvalue").
  ExtraVar("yourvar", 123).
  ...
  Send()
```

cmd/autoheal/object_template.go
zgalor: sure

cmd/autoheal/object_template.go
```diff
@@ -121,33 +121,55 @@ func (t *ObjectTemplate) Process(object interface{}, data interface{}) error {
 	if glog.V(2) {
 		glog.Infof("Data: %v", data)
 	}
-	return t.processValue(reflect.ValueOf(object), data)
+	v, err := t.processValue(reflect.ValueOf(object), data)
+	if err == nil {
+		object = v.Interface()
```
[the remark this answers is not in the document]
zgalor: Are you sure?
From what I tried this seems to work.
How else can I make sure `object` is updated?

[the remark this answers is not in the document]
jhernand: In this context `object` is a local variable of the `Process` method. If you assign something to it you are just changing the local variable, which won't have any effect in the original one, because it is passed by value. If you want to return a new result then you will have to use a return value, or else pass a pointer to the interface:

```go
func (t *ObjectTemplate) Process(input interface{}, data interface{}) (output interface{}, err) {
  ...
  output = ...
}
```

Or, with a pointer:

```go
func (t *ObjectTemplate) Process(object *interface{}, data interface{}) error {
  ...
  *object = ...
}
```

The second alternative is very confusing, so I'd suggest the first.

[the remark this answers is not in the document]
jhernand: Yes, I think that adding that check is enough.

cmd/autoheal/object_template.go
```diff
@@ -118,6 +118,11 @@ type ObjectTemplate struct {
 // parameter.
 //
 func (t *ObjectTemplate) Process(object interface{}, data interface{}) error {
+	if reflect.ValueOf(object).Kind() != reflect.Ptr {
```
jhernand: As you are going to use the `kind` twice, maybe is worth using a local variable:

```go
kind := reflect.ValueOf(object).Kind()
if kind != reflect.Ptr {
  return fmt.Errorf("....", kind)
}
```

cmd/autoheal/object_template.go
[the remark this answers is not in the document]
jhernand: As you added the check, I think you can now remove this assignment.

vendor/github.com/moolitayer/awx-client-go/awx/job_template_launch_resource.go
zgalor: Done. https://github.com/moolitayer/awx-client-go/pull/5

zgalor:
Last commit presents required API changes after rebasing with latest API patch https://github.com/openshift/autoheal/pull/69

@jhernand Please review

zgalor:
@jhernand Done

jhernand:
@zgalor the actual changes look good, and I would like to merge them. But the updated to dependencies commit is too large. Are you sure you are using the latest version of `dep`? Please re-check that you are using version 0.4.1, then run `dep ensure` again. It should only update the AWX files, and maybe a couple of others, but the current patch modifies more than 9000 files.

zgalor:
@jhernand the version I am using is `devel`. I install `dep` from sources. Should I use it differently?

zgalor:
@jhernand this should be ok now

PR #74, opened 2018-04-27
Update to kubernetes 1.10

jhernand:
This patch updates the `vendor` directory after the recent changes to the `Gopkg.toml` file.


jhernand:
This depends on #73, should be rebased once that is merged.

PR #76, opened 2018-04-29
Add alert content as an additional extra var to awx job

zgalor:
This resolves issue https://github.com/openshift/autoheal/issues/39

This PR completes and depends on https://github.com/openshift/autoheal/pull/67.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,10 @@ The `extraVars` parameter is optional, and if specified it is used to
 pass additional variables to the playbook, like with the `--extra-vars`
 option of the `ansible-playbook` command.
 
+Regardless to the `extraVars` setting, the content of the alert that 
+triggered the AWX job will be passed to the playbook as part of 
+`extraVars`, in a variable named `alert`.
+
 > Note that in order to be able to use this `extraVars` mechanism the
 > AWX job template should have the _Prompt on lauch_ box checked,
 > otherwise the variables passed will be ignored.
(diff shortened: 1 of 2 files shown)


zgalor:
@jhernand @moolitayer @nimrodshn @yaacov @zeari Please review (only last commit is relevant)

cmd/autoheal/awx_job.go
```diff
@@ -96,12 +96,14 @@ func (h *Healer) launchAWXJob(
...
 	action *autoheal.AWXJobAction,
 	rule *autoheal.HealingRule,
+	alert *alertmanager.Alert,
 ) error {
 	templateId := template.Id()
 	templateName := template.Name()
 	launchResource := connection.JobTemplates().Id(templateId).Launch()
 	response, err := launchResource.Post().
 		ExtraVars(action.ExtraVars).
+		ExtraVar("alerts", alert).
```
jhernand: If we are calling this `alerts` then it should be an array of alerts. The idea of that is that in the future we may want to have rules that are activated only when multiple alerts are fired, and then this would contain all those alerts. If it is going to be a single alert, which is fine as well, then I'd suggest to name it `alert`.

zgalor:
@jhernand Done

README.md
```diff
@@ -239,6 +239,9 @@ The `template` parameter is the name of the AWX job template.
 The `extraVars` parameter is optional, and if specified it is used to
 pass additional variables to the playbook, like with the `--extra-vars`
 option of the `ansible-playbook` command.
+Regardless to the `extraVars` setting, the alert that triggered the AWX
```
jhernand: I think this should be a separate paragraph. So can you add a leading blank line and end the paragraph with a dot?

zgalor:
@jhernand Done

Issue #14, opened 2018-03-27
Check correctly errors when checking if files exist


jhernand:
Currently the code that checks if files exist doesn't handle correctly all the errors potentially returned by `os.Stat`. In particular the code that checks if the Kubernetes client configuration file exist doesn't take into account that there may be errors other than the file not existing. See here:

https://github.com/openshift/autoheal/blob/4c067d65405d754339b5e173f15b01f891ad6624/cmd/autoheal/main.go#L69-L92

That should be fixed so that it checks and reports all potential errors.

PR #78, opened 2018-04-29
Cover more config reading errors

zeari:
Fixes https://github.com/openshift/autoheal/issues/14
Cover more config reading errors such as persmission denied on `.kube/config`.

```Bash
$ chmod 000 ~/.kube/config
$ ./_output/local/bin/linux/amd64/autoheal server --config-file=./examples/autoheal-dev.yml --logtostderr
I0429 15:56:39.455527   29602 server.go:96] The Kubernetes configuration file '.../.kube/config' can not be read: 'open .../.kube/config: permission denied', will try to use the in-cluster configuration
```

@jhernand @zgalor @elad661 Please review

### DIFF
--- a/cmd/autoheal/server.go
+++ b/cmd/autoheal/server.go
@@ -80,28 +80,30 @@ func serverRun(cmd *cobra.Command, args []string) {
 
 	// Load the Kubernetes configuration:
 	var config *rest.Config
-	_, err = os.Stat(serverKubeConfig)
-	if os.IsNotExist(err) {
-		glog.Infof(
-			"The Kubernetes configuration file '%s' doesn't exist, will try to use the "+
-				"in-cluster configuration",
-			serverKubeConfig,
-		)
-		config, err = rest.InClusterConfig()
-		if err != nil {
-			glog.Fatalf(
-				"Error loading in-cluster REST client configuration: %s",
-				err.Error(),
-			)
-		}
+	config, err = rest.InClusterConfig()
+	if err == nil {
+		glog.Infof("Using in-cluster configuration")
 	} else {
-		config, err = clientcmd.BuildConfigFromFlags(serverKubeAddress, serverKubeConfig)
-		if err != nil {
+		glog.Infof(
+			"Error loading in-cluster REST client configuration: %s. Trying kube config...",
+			err.Error(),
+		)
+		_, err = os.Stat(serverKubeConfig)
+		if os.IsNotExist(err) || os.IsPermission(err) || os.IsTimeout(err) {
 			glog.Fatalf(
-				"Error loading REST client configuration from file '%s': %s",
+				"The Kubernetes configuration file %s can not be read: ",
 				serverKubeConfig,
 				err.Error(),
 			)
+		} else {
+			config, err = clientcmd.BuildConfigFromFlags(serverKubeAddress, serverKubeConfig)
+			if err != nil {
+				glog.Fatalf(
+					"Error loading REST client configuration from file '%s': %s. No viable configuration found.",
+					serverKubeConfig,
+					err.Error(),
+				)
+			}
 		}
 	}
 


cmd/autoheal/server.go
```diff
@@ -81,25 +81,28 @@ func serverRun(cmd *cobra.Command, args []string) {
...
-		glog.Infof(
-			"The Kubernetes configuration file '%s' doesn't exist, will try to use the "+
-				"in-cluster configuration",
-			serverKubeConfig,
-		)
-		config, err = rest.InClusterConfig()
+	file, err := os.Open(serverKubeConfig) // For read access.
+
+	if err == nil {
+		file.Close()
```
zeari: defer file.Close() instead?

cmd/autoheal/server.go
```diff
@@ -81,25 +81,28 @@ func serverRun(cmd *cobra.Command, args []string) {
...
 	var config *rest.Config
 	_, err = os.Stat(serverKubeConfig)
-	if os.IsNotExist(err) {
-		glog.Infof(
-			"The Kubernetes configuration file '%s' doesn't exist, will try to use the "+
-				"in-cluster configuration",
-			serverKubeConfig,
-		)
-		config, err = rest.InClusterConfig()
+	file, err := os.Open(serverKubeConfig) // For read access.
```
yaacov: I think that using all the built-in functions ( like we used `IsNotExist` before ), and adding a take all `err != nil` at the end is nicer  ... but not a blocker :-)
```
func IsNotExist(err error) bool
func IsPermission(err error) bool
func IsTimeout(err error) bool
```

cmd/autoheal/server.go
cben: A way I've seen used to detect running under kubernetes is checking for `KUBERNETES_SERVICE_HOST` env var.
This var (plus _PORT) is currently what the `InClusterConfig` in the client uses to find the API:
https://github.com/kubernetes/client-go/blob/989be4278f353e42f26c416c53757d16fcff77db/rest/config.go#L313-L314

As discussed on https://github.com/kubernetes/kubernetes/issues/40973, this var has 2 problems:
1. It can be shadowed by a service that happens to be called `kubernetes` in current namespace.  (Not a problem for us, we control what's running in our namespace.)
2. it's sometimes unset due to startup race condition.  This being a tool for healing problems, we don't want to fail when problems happen :-)  A possibly better approach to find API is DNS `kubernetes.default.svc`; for now official clients were not changed (https://github.com/kubernetes/kubernetes/pull/43807 was rejected), but we can opt-in by explicitly setting `KUBERNETES_SERVICE_HOST` to `kubernetes.default.svc` in the pod config.

In any case, for the decision we can rely on whatever InClusterConfig will doing.  (If we're in a pod and it fails, we're hosed anyway.  But if we exit with error, kubelet can restart us, with backoff, which is not bad.)
=> I propose trying InClusterConfig first — it returns error if it doesn't find things it expects to have in a pod! — and then falling back to `~/.kube/config`.

cmd/autoheal/server.go
```diff
@@ -80,28 +80,30 @@ func serverRun(cmd *cobra.Command, args []string) {
...
-			)
-		}
-	} else {
-		config, err = clientcmd.BuildConfigFromFlags(serverKubeAddress, serverKubeConfig)
-		if err != nil {
+		_, err = os.Stat(serverKubeConfig)
+		if os.IsNotExist(err) || os.IsPermission(err) || os.IsTimeout(err) {
 			glog.Fatalf(
-				"Error loading REST client configuration from file '%s': %s",
+				"The Kubernetes configuration file %s can not be read: ",
```
moolitayer: There are two format arguments but only one % symbol referencing them

moolitayer:
@cben @zeari the previous code had the benefit of allowing the user to override the in cluster configuration (currently `kubeconfig` is ignored in cluster). Can we Attempt the in cluster config only if `kubeconfig` isn't specified? (very similar to current logic)

cben:
@moolitayer if that's a goal, then certainly my suggestion was had.
What's the use case for running in a cluster but overriding config?

moolitayer:
Do we always heal the cluster we run on? I'm not sure.
I think respecting what the user requests is straight forward and is most powerful.

This isn't critical to me @cben. Feel free to add the review labels and we can move this forward.

Issue #9, opened 2018-03-26
Send logs to stdout or stderr by default


jhernand:
Currently the `autoheal` application doesn't send the logs
to `stderr` or `stdout` by default. To do so it requires the
`-logalsotostderr` command line option. We should change
that so that logs go to `stderr` and `stdout` by default, without
requiring any option.

yaacov:
we do not have a `-logalsotostderr` ? 
does this refer to `-logtostderr` ?

yaacov:
Taking this one, didn't understand it it before :-)

yaacov:
@zeari is realy taking this one, but his not on openshift assignees yet.

PR #83, opened 2018-04-30
Add --logtostderr by default

zeari:
Fixes https://github.com/openshift/autoheal/issues/9
Add the flag --logtostderr to autoheal by default

@yaacov @jhernand Please review

### DIFF
--- a/cmd/autoheal/main.go
+++ b/cmd/autoheal/main.go
@@ -33,6 +33,7 @@ var rootCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(serverCmd)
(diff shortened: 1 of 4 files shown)


cmd/autoheal/main.go
```diff
@@ -33,6 +33,7 @@ var rootCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(serverCmd)
+	flag.Set("logtostderr", "true")
```
zeari: This seems to be the way the k8s vendor code does it. A quick search for `logtostderr` will show this line in multiple places.

Issue #44, opened 2018-04-11
Add tests 


nimrodshn:
Currently different parts of autoheal are not tested. (`rules_woker`, `config`, files under `pkg/` etc.), we should add tests for these.

cc: @jhernand

yaacov:
closing @jhernand ?

PR #88, opened 2018-04-30
object_template: Add tests

elad661:
Slowly but steadily increasing our test coverage while learning Go in the process.

This PR adds tests to object_template. Note that as object_template will support more types, more tests for these should be added.

cc: @jhernand @nimrodshn @zgalor @moolitayer Please review. All feedback is highly appreciated (even tiny nitpicks).

for reference: issue #44

### DIFF
--- a/cmd/autoheal/object_template_test.go
+++ b/cmd/autoheal/object_template_test.go
@@ -1,0 +1,69 @@
+/*
+Copyright (c) 2018 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"testing"
+)
+
+type TemplateTestDataNested struct {
+	StructKey string
+}
+
+type TemplateTestData struct {
+	Map    map[string]string
+	Str    string
+	Struct TemplateTestDataNested
+}
+
+func TestProcess(t *testing.T) {
+	template, err := NewObjectTemplateBuilder().
+		Delimiters("[", "]").
+		Variable("map", ".Map").
+		Variable("str", ".Str").
+		Variable("struct", ".Struct").
+		Build()
+
+	if err != nil {
+		t.Errorf("Error building ObjectTemplate: %v", err)
+	}
+
+	params := TemplateTestData{
+		Map:    map[string]string{"mapkey": "mapvalue"},
+		Str:    "This is a string",
+		Struct: TemplateTestDataNested{StructKey: "foo"},
+	}
+
+	input := map[string]string{
+		"a": "Test [ $foo ] test [ $bar ]",
+		"b": "Map=[ $map ], str=[ $str ], struct=[ $struct ]",
+	}
+	err = template.Process(&input, params)
+	if err != nil {
+		t.Errorf("Error processing template: %v", err)
+	}
+
+	if input["a"] != "Test [ $foo ] test [ $bar ]" {
+		t.Errorf("Input changed even though it didn't match template! %v", input["a"])
+	}
+
+	expected := "Map=map[mapkey:mapvalue], str=This is a string, struct={foo}"
+	if input["b"] != expected {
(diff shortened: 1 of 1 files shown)


cmd/autoheal/object_template_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+package main
+
+import (
+	"testing"
+)
+
+func TestNewObjectTemplateBuilder(t *testing.T) {
+	b := NewObjectTemplateBuilder()
+	if b.right != "}}" {
```
jhernand: I think it is better to avoid accessing private fields in the tests, as that means that we can't change the internals without also changing the tests. Instead of doing this you can try to process something that requires the delimiters, and then check that they did work. Same for variables or any other internal implementation detail: tests should not use them, in my opinion.

cmd/autoheal/object_template_test.go
```diff
@@ -0,0 +1,111 @@
...
+		t.Errorf("Expected }} right delimiter got %+v", b.right)
+	}
+	if b.left != "{{" {
+		t.Errorf("Expected {{ left delimiter got %+v", b.left)
+	}
+}
+
+func TestDelimiters(t *testing.T) {
+	b := ObjectTemplateBuilder{}
+	b.Delimiters("left", "right")
```
jhernand: Not sure if `left` and `right` are realistic delimiters. I'd suggest something that can be really useful, for exaple `[[` and `]]`. The process something that uses them, and check that it worked.

cmd/autoheal/object_template_test.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	if b.left != "left" {
+		t.Errorf("Expected left delimiter, got %+v", b.left)
+	}
+}
+
+func TestVariable(t *testing.T) {
+	b := ObjectTemplateBuilder{}
+	b.Variable("foo", "bar")
+	if b.variables["foo"] != "bar" {
```
jhernand: Same, instead of using the internal `variables` field, set the variable and then process something that requires it.

cmd/autoheal/object_template_test.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type TemplateTestData struct {
+	Map    map[string]string
+	Str    string
+	Struct TemplateTestDataNested
+}
+
+func TestProcess(t *testing.T) {
+	template, err := NewObjectTemplateBuilder().
```
jhernand: This is a very good example of the kind of test that I think we should be doing.

cmd/autoheal/object_template_test.go
```diff
@@ -0,0 +1,111 @@
...
+		Delimiters("[", "]").
+		Variable("map", ".Map").
+		Variable("str", ".Str").
+		Variable("struct", ".Struct").
+		Build()
```
jhernand: Should we check the value of `err` and fail the test if it isn't `nil`?

elad661: Done

elad661:
The new version of this PR removes all the tests that were done on private fields, and keeps only the tests for doing the actual processing. Tests for different input types should be added when support is added for these input types.

cmd/autoheal/object_template_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if err != nil {
+		t.Errorf("Error processing template: %+v", err)
```
moolitayer: `"Error building ObjectTemplate..."`

cmd/autoheal/object_template_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	params := TemplateTestData{
+		Map:    map[string]string{"mapkey": "mapvalue"},
+		Str:    "This is a string",
+		Struct: TemplateTestDataNested{StructKey: "foo"},
+	}
+
+	input := map[string]string{"a": "Test [ $foo ] test [ $bar ]",
+		"b": "Map=[ $map ], str=[ $str ], struct=[ $struct ]"}
```
moolitayer: Styling, can you please format this way:
```
	input := map[string]string{
		"a": "Test [ $foo ] test [ $bar ]",
		"b": "Map=[ $map ], str=[ $str ], struct=[ $struct ]",
	}
```

cmd/autoheal/object_template_test.go
```diff
@@ -0,0 +1,67 @@
...
+	err = template.Process(&input, params)
+	if err != nil {
+		t.Errorf("Error processing template: %+v", err)
```
moolitayer: You can omit the + from the format, it only has meaning if v is a struct

elad661:
thanks @moolitayer. I addressed the review feedback.

moolitayer:
@nimrodshn @zgalor @yaacov some additional please review

Issue #89, opened 2018-04-30
Verify AWX `PROMPT ON LAUNCH` option before running


zgalor:
In order to support passing `limit` and `extraVars` to AWX template, the `PROMPT ON LAUNCH` option must be checked.

We should verify it is checked for each of the fields in 2 stages:
1. When the rule is processed
2. Before the template is launched

If one of those fields is defined in the rule, but not set to `PROMPT ON LAUNCH` in the AWX template, we should log a warning to the user.

moolitayer:
Can you elaborate a bit more on what `PROMPT ON LAUNCH` is @zgalor ?

moolitayer:
OK I see https://github.com/openshift/autoheal/pull/90/files#diff-04c6e90faac2675aa89e2176d2eec7d8L247

> 1. When the rule is processed

What if awx is down while we are parsing configuration, do we want to abort loading it? will we retry?

I think writing a clear error message as you suggested in 2. is best.

PR #90, opened 2018-04-30
Add support for passing the `limit` parameter to AWX jobs

### LINKED ISSUE #45: Add support for passing the `limit` parameter to AWX jobs
When an AWX job is started it is possible to pass a `limit` parameter to limit the hosts where it will run. Currently the auto-heal service doesn't support this, but it should allow to pass it. For example, to pass the value of the `instance` label of the alert:

```yaml
awxJob:
  template: "My template"
  extraVars:
    myvar: myvalue
    yourvar: yourvalue
  limit: {{ $labels.instance }}
```

To be able to use that `limit` parameter the job template has to be configured with the _Prompt on launch_ flag, so we should also check that it is set.

zgalor:
resolve #45 

Note that templating the `limit` field will not work before the fix in https://github.com/openshift/autoheal/pull/95 is merged

Validating `PROMPT ON LAUNCH` configuration is also important so I opened https://github.com/openshift/autoheal/issues/89 to track that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,9 +241,17 @@ Regardless to the `extraVars` setting, the content of the alert that
 triggered the AWX job will be passed to the playbook as part of 
 `extraVars`, in a variable named `alert`.
 
-> Note that in order to be able to use this `extraVars` mechanism the
-> AWX job template should have the _Prompt on lauch_ box checked,
-> otherwise the variables passed will be ignored.
+The `limit` parameter is optional, and if specified it is passed to
+AWX to constrain the list of hosts managed or affected by the 
+playbook. Multiple patterns can be separated by colons (`:`). 
+As with core Ansible, `a:b` means "in group a or b", `a:b:&c` means 
+"in a or b but must be in c", and `a:!b` means "in a, and definitely 
+not in b".
+
+> Note that in order to be able to use `extraVars` and `limit`
+> mechanisms the AWX job template should have the 
+> _Prompt on lauch_ box checked, otherwise the variables passed 
+> will be ignored.
 
 The values of all the parameters inside `awxJob` are processed as [Go
 templates](https://golang.org/pkg/text/template) before executing the
@@ -265,6 +273,14 @@ awxJob:
   template: "My template"
   extraVars: 
     node: "{{ $labels.node }}"
+```
+
+Limit execution to a host, calculated from the `instance` label:
+
+```yaml
+awxJob:
+  template: "My template"
+  limit: "{{ $labels.instance }}"
 ```
 
 ### Alertmanager Configuration
--- a/pkg/apis/autoheal/types.go
+++ b/pkg/apis/autoheal/types.go
@@ -84,6 +84,11 @@ type AWXJobAction struct {
 	// ExtraVars are the extra variables that will be passed to job.
 	// +optional
 	ExtraVars JsonDoc
+
+	// Limit is a pattern that will be passed to the job to constrain
+	// the hosts that will be affected by the playbook.
+	// +optional
+	Limit string
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/apis/autoheal/v1alpha2/types.go
+++ b/pkg/apis/autoheal/v1alpha2/types.go
@@ -84,6 +84,11 @@ type AWXJobAction struct {
 	// ExtraVars are the extra variables that will be passed to job.
 	// +optional
 	ExtraVars JsonDoc `json:"extraVars,omitempty"`
+
+	// Limit is a pattern that will be passed to the job to constrain
+	// the hosts that will be affected by the playbook.
+	// +optional
+	Limit string `json:"limit,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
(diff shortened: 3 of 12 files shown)


pkg/apis/autoheal/types.go
```diff
@@ -84,6 +84,10 @@ type AWXJobAction struct {
 	// ExtraVars are the extra variables that will be passed to job.
 	// +optional
 	ExtraVars JsonDoc
+
+	// Limit in the execution limit that will be passed to job.
```
jhernand: I don't quite understand this comment, sorry :-) , can you reword it?

pkg/apis/autoheal/v1alpha2/types.go
```diff
@@ -84,6 +84,10 @@ type AWXJobAction struct {
 	// ExtraVars are the extra variables that will be passed to job.
 	// +optional
 	ExtraVars JsonDoc `json:"extraVars,omitempty"`
+
+	// Limit in the execution limit that will be passed to job.
```
jhernand: Same here.

pkg/apis/autoheal/types.go
yaacov: **Limit**: A host pattern to further constrain the list of hosts managed or affected by the playbook. Multiple patterns can be separated by colons (`:`). As with core Ansible, `a:b` means `in group a or b`, `a:b:&c` means `in a or b but must be in c`, and `a:!b` means `in a, and definitely not in b`.

http://docs.ansible.com/ansible-tower/latest/html/userguide/job_templates.html

zgalor:
@jhernand I improved the documentation, please continue reviewing

README.md
```diff
@@ -241,9 +241,17 @@ Regardless to the `extraVars` setting, the content of the alert that
...
 
-> Note that in order to be able to use this `extraVars` mechanism the
-> AWX job template should have the _Prompt on lauch_ box checked,
-> otherwise the variables passed will be ignored.
+The `limit` parameter is optional, and if specified it is passed to
+AWX to constrain the list of hosts managed or affected by the 
+playbook. Multiple patterns can be separated by colons (”:”). 
+As with core Ansible, “a:b” means “in group a or b”, “a:b:&c” means 
+“in a or b but must be in c”, and “a:!b” means “in a, and definitely 
+not in b”.
```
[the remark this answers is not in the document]
zgalor: That's what happens when you copy-paste with formatting... ;) Fixing

zgalor:
@jhernand Done

zgalor:
@jhernand any other comments?

moolitayer:
@zgalor Please bug me while @jhernand is on holiday. I'm reviewing.

README.md
```diff
@@ -241,9 +241,17 @@ Regardless to the `extraVars` setting, the content of the alert that
...
-> AWX job template should have the _Prompt on lauch_ box checked,
-> otherwise the variables passed will be ignored.
+The `limit` parameter is optional, and if specified it is passed to
+AWX to constrain the list of hosts managed or affected by the 
+playbook. Multiple patterns can be separated by colons (`:`). 
+As with core Ansible, `a:b` means "in group a or b", `a:b:&c` means 
+"in a or b but must be in c", and `a:!b` means "in a, and definitely 
+not in b".
+
+> Note that in order to be able to use this `extraVars` an `limit`
```
moolitayer: remove `this`

README.md
```diff
@@ -241,9 +241,17 @@ Regardless to the `extraVars` setting, the content of the alert that
...
+> mechanisms the AWX job template should have the 
+> _Prompt on lauch_ box checked, otherwise the variables passed
```
moolitayer: I know this wasn't introduced here, but why not PROMPT ON LAUNCH like it is in the UI?

README.md
```diff
@@ -241,9 +241,17 @@ Regardless to the `extraVars` setting, the content of the alert that
...
+AWX to constrain the list of hosts managed or affected by the 
+playbook. Multiple patterns can be separated by colons (`:`). 
+As with core Ansible, `a:b` means "in group a or b", `a:b:&c` means 
+"in a or b but must be in c", and `a:!b` means "in a, and definitely 
+not in b".
+
+> Note that in order to be able to use this `extraVars` an `limit`
+> mechanisms the AWX job template should have the 
+> _Prompt on lauch_ box checked, otherwise the variables passed 
+> will be ignored.
```
moolitayer: How about:
Note, if the PROMPT ON LAUNCH checkbox is not selected in the AWX template for `extraVars` or `limit`, they will not override the default defined there.

README.md
```diff
@@ -267,6 +275,14 @@ awxJob:
     node: "{{ $labels.node }}"
 ```
 
+Or to limit execution to a host, calculated from the `instance` label:
```
moolitayer: Can you remove the 'Or to'? can be written as:
Limit execution to the host calculated from the `instance` label:

If you think this isn't working with the paragraph above you can rephrase it, or leave as is.

README.md
moolitayer: Ignore the other suggestions for now. Fix this sentence to:
Note that in order to be able to use `extraVars` and `limit`

zgalor:
@moolitayer Done

moolitayer:
@zgalor now that #88  is merged, please add a test for the bug you fixed.

zgalor:
@moolitayer I opened https://github.com/openshift/autoheal/pull/95 for the bug fix, and I will add the tests there

PR #94, opened 2018-05-02
Adding tests for alert worker

nimrodshn:
cc: @yaacov @jhernand @moolitayer  PTAL

partially resolving issue #44

### DIFF
--- a/cmd/autoheal/alert_worker_test.go
+++ b/cmd/autoheal/alert_worker_test.go
@@ -1,0 +1,157 @@
+/*
+Copyright (c) 2018 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"path/filepath"
+	"reflect"
+	"testing"
+
+	"github.com/openshift/autoheal/pkg/alertmanager"
+	"github.com/openshift/autoheal/pkg/apis/autoheal"
+	batch "k8s.io/api/batch/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestPickAlert(t *testing.T) {
+	file := filepath.Join("..", "..", "testdata", "empty-config.yml")
+	healer, err := NewHealerBuilder().
+		ConfigFile(file).
+		Build()
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	alert := &alertmanager.Alert{
+		Status: "firing",
+		Labels: map[string]string{
+			"mylabel": "myvalue",
+		},
+	}
+
+	healer.alertsQueue.Add(alert)
+
+	// one alert in queue.
+	if healer.pickAlert() != true {
+		t.Errorf("Expected pickAlert to return true (i.e. there is an alert in alertQueue) but got false")
+	}
+}
+
+func TestStartHealingAWXJob(t *testing.T) {
+	file := filepath.Join("..", "..", "testdata", "empty-config.yml")
+	healer, err := NewHealerBuilder().
+		ConfigFile(file).
+		Build()
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	actionRunner := FakeActionRunner{
+		RuleAlertMap: make(map[string]*alertmanager.Alert),
+	}
+
+	healer.actionRunners[ActionRunnerTypeAWX] = actionRunner
+
+	alert := &alertmanager.Alert{
+		Status: "firing",
+		Labels: map[string]string{
+			"mylabel": "myvalue",
+		},
+	}
+
+	rule := &autoheal.HealingRule{
+		ObjectMeta: meta.ObjectMeta{
+			Name: "test-rule",
+		},
+		Labels: map[string]string{
+			"mylabel": "myvalue",
+		},
+		AWXJob: &autoheal.AWXJobAction{
+			Template: "Test AWX JOB",
+		},
+	}
+
+	// Add the rule to rulesCache
+	healer.rulesCache.Store(rule.ObjectMeta.Name, rule)
+
+	healer.startHealing(alert)
+
+	expected := map[string]*alertmanager.Alert{
+		rule.ObjectMeta.Name: alert,
+	}
+
+	if reflect.DeepEqual(expected, actionRunner.RuleAlertMap) != true {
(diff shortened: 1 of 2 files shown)


cmd/autoheal/alert_worker_test.go
```diff
@@ -0,0 +1,149 @@
...
+		Labels: map[string]string{
+			"mylabel": "myvalue",
+		},
+	}
+
+	healer.alertsQueue.Add(alert)
+
+	// one alert in queue.
+	if healer.pickAlert() != true {
+		t.Errorf("Expected pickAlert to return true (i.e. there is an alert in alertQueue) but got false")
```
elad661: should we added a test for returning false when the queue is empty?

cmd/autoheal/alert_worker_test.go
```diff
@@ -0,0 +1,149 @@
...
+	file := filepath.Join("..", "..", "testdata", "empty-config.yml")
+	healer, err := NewHealerBuilder().
+		ConfigFile(file).
+		Build()
+
+	healer.actionRunners[ActionRunnerTypeAWX] = FakeAWXActionRunner{}
+	healer.actionRunners[ActionRunnerTypeBatch] = FakeBatchActionRunner{}
+
+	if err != nil {
+		t.Error(err)
```
elad661: `t.Errorf("Error building healer: %v", err`

And this probably should be moved to be right after building, before assigning the action runners

cmd/autoheal/alert_worker_test.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	// Test 0
```
elad661: I feel like this comment should be more descriptive

cmd/autoheal/alert_worker_test.go
```diff
@@ -0,0 +1,149 @@
...
+	expected := fmt.Sprintf("Fake AWXActionRunner called with rule '%s' and alert '%s'",
+		change0.Rule.ObjectMeta.Name,
+		alert0.Name(),
+	)
+
+	if output.Error() != expected {
+		t.Errorf("Expected to get error: %s, instead got %s", expected, output.Error())
+	}
+
+	// Test 1
```
elad661: I think, but I'm not sure, that you can separate these tests to two functions. Otherwise the failure of the first test might hide another failure of the second test - but I'm not 100% sure on the standard practice for this in Go, so don't take my word for it

cmd/autoheal/alert_worker_test.go
```diff
@@ -0,0 +1,149 @@
...
+			},
+		},
+	}
+
+	// Add the rule change to rulesCache
+	healer.processRuleChange(change0)
+
+	output := healer.startHealing(alert0)
+
+	expected := fmt.Sprintf("Fake AWXActionRunner called with rule '%s' and alert '%s'",
```
elad661: Maybe also test if it was called with the correct AWXJob? And add another "dummy" rule to make sure rule matching works as intended?

cmd/autoheal/alert_worker_test.go
nimrodshn: @elad661 Can't really return `false` because the work implementation we are currently using blocks on `Get` method. (But a valid suggestion!)

cmd/autoheal/alert_worker_test.go
nimrodshn: `healer_test.go` contains a bunch of rule matching tests so I didn't want to touch on that logic.
Also, I feel like this essentially does check that it is the correct rule (If you look at the`RunAction` implementation in the stubs `Fake*ActionRunner` at `healer_test.go`).

cmd/autoheal/alert_worker_test.go
nimrodshn: Hmm we could use a test table (see [this](https://nathanleclaire.com/blog/2015/10/10/interfaces-and-composition-for-effective-unit-testing-in-golang/) article on testing in GO.) but it's a bit of an over kill for two tests.

cmd/autoheal/alert_worker_test.go
nimrodshn: @yaacov What do you think?

cmd/autoheal/alert_worker_test.go
nimrodshn: Fixed :+1: Thanks mate.

cmd/autoheal/alert_worker_test.go
```diff
@@ -0,0 +1,149 @@
...
+	file := filepath.Join("..", "..", "testdata", "empty-config.yml")
+	healer, err := NewHealerBuilder().
+		ConfigFile(file).
+		Build()
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	defer runtime.HandleCrash()
```
moolitayer: Why do we need this?

nimrodshn: Sorry, that slipped in there.

nimrodshn:
@yaacov @moolitayer PTAL

moolitayer:
Nimrod please split test 0 and test 1 to individual tests.

Also this test is testing functionality of rules_worker. Can you create a new rules_worker test? In alert worker test I think setting rulesCache is best.

nimrodshn:
> Can you create a new rules_worker test? In alert worker test I think setting rulesCache is best.
Sure will create a new `rules_worker` test! What do you mean by you're second comment?

moolitayer:
The comment about setting rulesCache? I mean currently you are calling processRuleChange to update the rules for your test. Since that part is actually testing rules_worker in this test you will have to find some other way.

nimrodshn:
> Since that part is actually testing rules_worker in this test you will have to find some other way.

What if I add a test for `processRuleChange` in the new `rules_worker` test? :smile:

nimrodshn:
@moolitayer done. working on `rules_worker` :+1:

cmd/autoheal/healer_test.go
```diff
@@ -570,9 +571,19 @@ func makeHealer(t *testing.T, name string) *Healer {
 type FakeAWXActionRunner struct{}
 
 func (f FakeAWXActionRunner) RunAction(rule *autoheal.HealingRule, action interface{}, alert *alertmanager.Alert) error {
-	glog.Infof("Fake AWXActionRunner called with rule '%s' and alert '%s'",
+	errMsg := fmt.Sprintf("Fake AWXActionRunner called with rule '%s' and alert '%s'",
 		rule.ObjectMeta.Name,
 		alert.Name(),
 	)
-	return nil
+	return errors.New(errMsg)
```
moolitayer: I don't like the Fake runner returning error when everything goes well.
You can have save list of rules and alerts a specific runner was asked to run and after using it you can test the list to container what you expect.

nimrodshn:
@moolitayer fixed, PTAL :smile_cat:

moolitayer:
Looking good. I have two minor comments:
I think we can we use a regular map and not a sync one. Please compare the result by comparing two maps. Expected vs actual. I think reflect.DeepEqual(map1, map2) is a good way to do that.

moolitayer:
Thanks @nimrodshn!

PR #95, opened 2018-05-02
Fix bug of templating a string field

zgalor:
This was presented in the refactoring done in e7b7f1ce24be397c12762b72404dbdabbcf0aa36
String values have to be set explicitly using `SetString()`
`SetString` is called only for settable values.
See laws of reflection here:
https://blog.golang.org/laws-of-reflection
    
Also added tests for `Process()` with the inputs struct and basic string


moolitayer:
Thanks @zgalor, needs test

PR #97, opened 2018-05-07
Buffer file change events

yaacov:
**Description**

Last step in solving #41

When editing config files using some editors and automatic tools we may get several file change events in a small time frame. This PR buffer events that happen during 1 sec into one event.

**Screenshoots**

Before:
before-1

After:
after-1

### DIFF
--- a/pkg/config/event_listener.go
+++ b/pkg/config/event_listener.go
@@ -19,6 +19,8 @@ limitations under the License.
 package config
 
 import (
+	"time"
+
 	"github.com/golang/glog"
 	"github.com/yaacov/observer/observer"
 )
@@ -64,6 +66,9 @@ func (e *eventListener) open() {
 	if e.configFilesChangedObserver == nil {
 		e.configFilesChangedObserver = new(observer.Observer)
 		e.configFilesChangedObserver.Open()
+
+		// Buffer file change events, 1e6 ns == 1 ms
+		e.configFilesChangedObserver.SetBufferDuration(1 * time.Millisecond)
 	}
 
 	// Start a change watcher over loaded config files.


yaacov:

pkg/config/event_listener.go
```diff
@@ -64,6 +66,9 @@ func (e *eventListener) open() {
 	if e.configFilesChangedObserver == nil {
 		e.configFilesChangedObserver = new(observer.Observer)
 		e.configFilesChangedObserver.Open()
+
+		// Buffer file change events
+		e.configFilesChangedObserver.SetBufferDuration(1 * time.Second)
```
jhernand: What is the reason for the multiple events? Multiple writes by the the program (the editor) that is changing the configuration? Or multiple actions by the user of that program?

Does this mean that any change to the configuration will always be delayed at least one second? If there is only one event will it also be delayed? Or only the second and subsequent events will be delayed?

yaacov: > What is the reason for the multiple events?

a. Some editors, do `copy to tmp` -> `edit tmp` -> `move tmp over original`. Because of a limitation with `fsnotify` that will trigger two `write` events for the 'delete' of the old config file and for the 'creation/move' of the new one (I think `fsnotify` use lower denominator of all it's possible backends). 

``` bash
vim a.txt
# only one :w was done ...
```

``` bash
14:35 $ while true; do inotifywait ./ ; done
Setting up watches.
Watches established.
./ CREATE 4913
Setting up watches.
Watches established.
./ CREATE a.txt
Setting up watches.
Watches established.
./ ATTRIB a.txt
Setting up watches.
Watches established.
./ CLOSE_WRITE,CLOSE .a.txt.swp
Setting up watches.
Watches established.
./ OPEN,ISDIR 
Setting up watches.
Watches established.
./ OPEN README.md
Setting up watches.
Watches established.

```

b. Automation may edit multiple config files, and we want to act only when the last one is edited.

> Multiple writes by the the program (the editor) that is changing the configuration? Or multiple actions by the user of that program?

We want to catch both cases, if the editor/automation does `delete/move` we only want to act on the last `move`, and if the automation edit multiple files, we want to act only when the last file is edited.

> Does this mean that any change to the configuration will always be delayed at least one second?

Yes, it will wait to try and catch events that may follow.

> If there is only one event will it also be delayed? 

Yes, we will wait to see if more events come.

> Or only the second and subsequent events will be delayed?

No, second events are not delayed.

Note:
``` python
while true:
    if len(events) == 1:
        wait for N sec # continue collecting events
        emit all events
```

yaacov: @jhernand 

Better command line to see the problem (onle done one :w on vim):

``` bash
inotifywait -mr ./
Setting up watches.  Beware: since -r was given, this may take a while!
Watches established.
./ CREATE 4913
./ OPEN 4913
./ ATTRIB 4913
./ CLOSE_WRITE,CLOSE 4913
./ DELETE 4913
./ MOVED_FROM a.txt
./ MOVED_TO a.txt~
./ CREATE a.txt
./ OPEN a.txt
./ MODIFY a.txt
./ MODIFY a.txt
./ ATTRIB a.txt
./ CLOSE_WRITE,CLOSE a.txt
./ ATTRIB a.txt
./ DELETE a.txt~
./ MODIFY .a.txt.swp

```

jhernand: I am not against doing this, rather the opposite, but 1 seconds is a lot of time for a computer. Can we make it shorter, 1 ms, for example? That should be enough for any editor to do the move, save, move cycle.

yaacov: :+1: looking into it

pkg/config/event_listener.go
```diff
@@ -64,6 +66,9 @@ func (e *eventListener) open() {
 	if e.configFilesChangedObserver == nil {
 		e.configFilesChangedObserver = new(observer.Observer)
 		e.configFilesChangedObserver.Open()
+
+		// Buffer file change events, 1e6 ns == 1 ms
+		e.configFilesChangedObserver.SetBufferDuration(1000000 * time.Nanosecond)
```
jhernand: Any reason to not use `1 * time.Millisecond`?

yaacov: > Any reason to not use 1 * time.Millisecond?

@jhernand , No, playing with timings ... changed to `Millisecond` :+1: 

cc @nimrodshn ( who also asked the same question :-) )

jhernand: OK. It looks good to me as is. Should I merge?

yaacov: I'm ok with it :-)

PR #98, opened 2018-05-08
Improve consistency of log messages

jhernand:
This patch changes some of the log messages so that the general style is more consistent. In general a log message should be a grammatically correct sentence, starting with upper case and ending without punctuation. If values, like file names, are included then they should be enclosed in single quotes, so that it is
easy to spot unusual values like empty or blank strings:

```go
glog.Infof("Loading configuration from file '%s'", file)
```

When an error needs to be reported in the log message it should be included at the end of the message, after a colon:

```go
glog.Errorf("Can't load configuration file '%s': %s", file, err)
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -97,22 +97,24 @@ func (c *Config) watch() {
 	configFiles := c.configFiles()
 	err := e.configFilesChangedObserver.Watch(configFiles)
 	if err != nil {
-		glog.Errorf("watch [Error]: %v", err)
+		glog.Errorf("Can't watch configuration files: %s", err)
+		return
 	}
-	glog.Infof("watch: %v", configFiles)
+	for _, file := range configFiles {
+		glog.Infof("Watching configuration file '%s'", file)
+	}
 
 	// Load new configuration when config files change.
-	e.configFilesChangedObserver.AddListener(func(event interface{}) {
+	e.configFilesChangedObserver.AddListener(func(_ interface{}) {
 		// Lock this function
 		c.loadMutex.Lock()
 		defer c.loadMutex.Unlock()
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -97,22 +97,24 @@ func (c *Config) watch() {
 	configFiles := c.configFiles()
 	err := e.configFilesChangedObserver.Watch(configFiles)
 	if err != nil {
-		glog.Errorf("watch [Error]: %v", err)
+		glog.Error("Can't watch configuration files: %s", err)
```
yaacov: '%s' ?

jhernand: I think that error messages should be included without quotes, as they may contain quotes themselves.

pkg/config/config.go
```diff
@@ -97,22 +97,24 @@ func (c *Config) watch() {
...
+		return
+	}
+	for _, file := range configFiles {
+		glog.Infof("Watching configuration file '%s'", file)
 	}
-	glog.Infof("watch: %v", configFiles)
 
 	// Load new configuration when config files change.
 	e.configFilesChangedObserver.AddListener(func(event interface{}) {
```
yaacov: if we do not use the event var, we can use some shorter name, or _ ?

pkg/config/config.go
yaacov: :+1:

pkg/config/config.go
jhernand: Yes, sure, I will change that.

yaacov:
/lgtm
:-)

yaacov:
/lgtm
:-) :-)

PR #99, opened 2018-05-08
Acquire mutex inside `Config.load`

jhernand:
Currently the mutext that prevents concurrent modification of the configuration structures is acquired before calling the `load` method, in two diffent places. This patch moves that to inside the `load` method, so that it will be only in one place.


yaacov:
/lgtm

:+1:  I moved the lock to the watch + listener because I wanted the `emit` in line 122 to run with the `load` in 155. But less locks are better ... moving to load makes sense to me, especially since watch is called only once.

PR #100, opened 2018-05-08
Check config load errors

jhernand:
Currently the errors that happen when initially loading the  configuration are ignored. For example, if the AWX credentials are  stored in a Kubernetes secret but the secret does not exist, or the connection to the API server does not work, the server will just continue using an empty user name and password. To avoid that this patch changes the code so that it explicitly checks for errors returned by the `load` method.


jhernand:
Note that this depends on #99.

yaacov:
/lgtm

:+1:

PR #103, opened 2018-05-09
Check more errors when loading kubconfiig

yaacov:
**Description**

When loading kubeconfig we only check for `IsNotExist` error, other errors will get generic error message. This PR give specific error message for each type of file error. 

**Terminal dumps**

File does not exist:
```
14:52 $ _output/local/bin/linux/amd64/autoheal server --kubeconfig examples/autoheal-dev

I0509 14:52:08.806540    4687 server.go:144] Info: The Kubernetes configuration file 'examples/autoheal-dev' doesn't exist
I0509 14:52:08.806674    4687 server.go:148] Try to use the in-cluster configuration
F0509 14:52:08.806696    4687 server.go:153] Error loading in-cluster REST client configuration: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
```

File is not a config file:
```
14:52 $ _output/local/bin/linux/amd64/autoheal server --kubeconfig examples/autoheal-dev.yml 

F0509 14:52:22.328359    4744 server.go:138] Error loading REST client configuration from file 'examples/autoheal-dev.yml': invalid configuration: no configuration has been provided
```

File is a directory:
```
14:52 $ _output/local/bin/linux/amd64/autoheal server --kubeconfig examples

F0509 14:52:35.216649    4802 server.go:157] Error: The Kubernetes configuration path 'examples' is a direcory
```

Default file is not exist:
```
14:52 $ mv ~/.kube/config ~/.kube/config.old
14:53 $ _output/local/bin/linux/amd64/autoheal server

I0509 14:53:19.247959    4986 server.go:144] Info: The Kubernetes configuration file '/home/yzamir/.kube/config' doesn't exist
I0509 14:53:19.248039    4986 server.go:148] Try to use the in-cluster configuration
F0509 14:53:19.248055    4986 server.go:153] Error loading in-cluster REST client configuration: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
```

Default file exist:
```
14:53 $ mv ~/.kube/config.old ~/.kube/config
14:53 $ _output/local/bin/linux/amd64/autoheal server

I0509 14:53:32.269368    5093 config.go:193] Loading configuration file 'autoheal.yml'
...
```

Fix: https://github.com/openshift/autoheal/issues/14
Reference: https://kubernetes-v1-4.github.io/docs/user-guide/kubectl/kubectl_config/

### DIFF
--- a/pkg/config/builder_test.go
+++ b/pkg/config/builder_test.go
@@ -57,6 +57,10 @@ func TestFiles(t *testing.T) {
 
 	l.Files([]string{file0.Name(), file1.Name()})
 	cfg, err := l.Build()
+	if err != nil {
+		t.Errorf("An error occured! %s", err)
+	}
+	defer cfg.ShutDown()
 
 	expected := &Config{
 		awx: &AWXConfig{
@@ -298,23 +302,26 @@ func TestLoadFile(t *testing.T) {
 	l.File(file.Name())
 
 	for _, test := range configsTest {
-		file.WriteAt([]byte(test.configString), 0)
-		cfg, err := l.Build()
-		if err != nil {
-			t.Errorf("An error occured! %s", err)
-		}
+		func() {
+			file.WriteAt([]byte(test.configString), 0)
+			cfg, err := l.Build()
+			if err != nil {
+				t.Errorf("An error occured! %s", err)
+			}
+			defer cfg.ShutDown()
 
-		if !reflect.DeepEqual(cfg.awx, test.expected.awx) {
-			t.Errorf("Expected %+v but got %+v", test.expected.awx, cfg.awx)
-		}
+			if !reflect.DeepEqual(cfg.awx, test.expected.awx) {
+				t.Errorf("Expected %+v but got %+v", test.expected.awx, cfg.awx)
+			}
 
-		if !reflect.DeepEqual(cfg.throttling, test.expected.throttling) {
-			t.Errorf("Expected %+v but got %+v", test.expected.throttling, cfg.throttling)
-		}
+			if !reflect.DeepEqual(cfg.throttling, test.expected.throttling) {
+				t.Errorf("Expected %+v but got %+v", test.expected.throttling, cfg.throttling)
+			}
 
-		if !reflect.DeepEqual(cfg.rules.rules, test.expected.rules.rules) {
-			t.Errorf("Expected %+v but got %+v", test.expected.rules.rules, cfg.rules.rules)
-		}
+			if !reflect.DeepEqual(cfg.rules.rules, test.expected.rules.rules) {
+				t.Errorf("Expected %+v but got %+v", test.expected.rules.rules, cfg.rules.rules)
+			}
+		}()
 	}
 }
 
@@ -374,10 +381,10 @@ func TestLoadDir(t *testing.T) {
 	l.File(dir)
 
 	cfg, err := l.Build()
-
 	if err != nil {
 		t.Errorf("An error occured! %s", err)
 	}
+	defer cfg.ShutDown()
 
 	if !reflect.DeepEqual(cfg.awx, expected.awx) {
 		t.Errorf("Expected %+v but got %+v", expected.awx, cfg.awx)
--- a/cmd/autoheal/server.go
+++ b/cmd/autoheal/server.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 
@@ -50,7 +51,7 @@ func init() {
 	serverFlags.StringVar(
 		&serverKubeConfig,
 		"kubeconfig",
-		filepath.Join(homedir.HomeDir(), ".kube", "config"),
+		"",
 		"Path to a Kubernetes client configuration file. Only required when running "+
 			"outside of a cluster.",
 	)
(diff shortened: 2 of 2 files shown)


yaacov:
@zeari @nimrodshn @zgalor @moolitayer @elad661 @jhernand . Please review.

stevekuznetsov:
Might not be _your_ data race, but you've got one in your unit tests :)

yaacov:
> Might not be your data race, but you've got one in your unit tests :)

@stevekuznetsov  lines 104 and 83:
```
github.com/openshift/autoheal/pkg/config.(*Config).load()
      github.com/openshift/autoheal/pkg/config/_test/_obj_test/config.go:104 +0x113
  github.com/openshift/autoheal/pkg/config.(*Config).watch.func1()
      github.com/openshift/autoheal/pkg/config/_test/_obj_test/config.go:83 +0x11d
```
does not exist in current code :-( what am I missing ?
https://github.com/openshift/autoheal/blob/master/pkg/config/config.go#L104
https://github.com/openshift/autoheal/blob/master/pkg/config/config.go#L83

yaacov:
@stevekuznetsov help :-)
``` bash
$ make check
...
ok  	github.com/openshift/autoheal/pkg/alertmanager	(cached)	coverage: 100.0% of statements
ok  	github.com/openshift/autoheal/pkg/apis/autoheal	(cached)	coverage: 8.0% of statements
ok  	github.com/openshift/autoheal/pkg/config	(cached)	coverage: 57.5% of statements
ok  	github.com/openshift/autoheal/pkg/memory	(cached)	coverage: 90.0% of statements
[INFO] hack/test-go.sh exited with code 0 after 00h 00m 06s
```
consistently works on my system, what test can I run to debug this ?

jhernand:
Not sure if it is related but I see this in the [raw log](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_autoheal/103/pull-autoheal-check/994225694009462784/build-log.txt) of the job:

```
/tmp/go-build213817817/github.com/openshift/autoheal/cmd/autoheal/_test/_obj_test/server.go:102:5: undefined: os.IsTimeout
FAIL	github.com/openshift/autoheal/cmd/autoheal [build failed]
```

That happens because the job uses Go 1.9 to build (like OpenShift itself) and the `os.IsTimeout` was introduced in Go 1.10.

@yaacov I'd suggest you address that, to avoid using `os.IsTimeout`, then we can check again if the data race appears.

By the way, to reproduce the job in your environment you can do the following:

```
docker run -ti -v /your-go-path/src/github.com/openshift/autoheal:/source openshift/origin-release:golang-1.9 make -C /source check
```

yaacov:
@stevekuznetsov @jhernand thanks, installed 1.9 and getting the error, will fix ...

jhernand:
@stevekuznetsov  I selected 1.0 because I see that the job to build OpenShift itself uses 1.9 as well, so I wanted to use the same. Is there a policy about what version of Go to use, or can each project decide independently?

@yaacov I tested without your patch and the race was already there (using Go 1.9, works well using Go 1.10). Would you be so kind to investigate it?

yaacov:
> Would you be so kind to investigate it?

@jhernand :+1:  On it, now :-)

jhernand:
@yaacov while checking this I added the `-work` option to the `go test` command, just to take a look at the temporary files that Go generates for the tests, and surprisingly I found that it doesn't fail. That `-work` option just tells Go to print the name of the temporary working directory, and to not delete it when it finishes. I think that what happens is that our observer keeps running after the test has finished, and then it detects that Go is removing the files, and triggers a reload, which may cause all sorts of problems. Can we make sure that we shutdown the loader before ending the tests to see if that solves the problem?

yaacov:
@jhernand :+1:   `cfg.ShutDown()` worked for me, whoo :fireworks: waiting to see what the ci tests say ...

stevekuznetsov:
@smarterclayton @jupierce do we have any guidance as to which version of Go to depend on for builds? My first blush impression would be to use the same one that Origin is, but I don't really know if we have a more specific recommendation.

pkg/config/builder_test.go
```diff
@@ -315,6 +317,8 @@ func TestLoadFile(t *testing.T) {
 		if !reflect.DeepEqual(cfg.rules.rules, test.expected.rules.rules) {
 			t.Errorf("Expected %+v but got %+v", test.expected.rules.rules, cfg.rules.rules)
 		}
+
+		cfg.ShutDown()
```
jhernand: This will not run if the test ends or panics before this point. Can you use `defer cfg.Shutdown()` instead?

pkg/config/builder_test.go
```diff
@@ -374,6 +378,7 @@ func TestLoadDir(t *testing.T) {
 	l.File(dir)
 
 	cfg, err := l.Build()
+	defer cfg.ShutDown()
```
jhernand: I think that this should go after checking `err`, as if `err` is not `nil` then the configuration shouldn't have been created, and `ShutDown()` should have no effect.

jhernand:
@stevekuznetsov I see that the job is not running automatically, I have to explcilty add `/test ...`. Is there anything else that we need to do so that it is run automatically with every pull request and with every change to existing pull requests?

pkg/config/builder_test.go
yaacov: :+1:

pkg/config/builder_test.go
yaacov: :+1:

pkg/config/builder_test.go
```diff
@@ -297,12 +302,18 @@ func TestLoadFile(t *testing.T) {
 	l := NewBuilder()
 	l.File(file.Name())
 
+	mutex := sync.Mutex{}
 	for _, test := range configsTest {
+		// Lock this loop
+		mutex.Lock()
+		mutex.Unlock()
```
jhernand: What is the effect of this? I'd say it has no effect.

pkg/config/builder_test.go
```diff
@@ -297,12 +302,18 @@ func TestLoadFile(t *testing.T) {
 	l := NewBuilder()
 	l.File(file.Name())
 
+	mutex := sync.Mutex{}
 	for _, test := range configsTest {
+		// Lock this loop
+		mutex.Lock()
+		defer mutex.Unlock()
```
jhernand: Now you need to explain me why we need to acquire the lock :-) .

yaacov: well we don't ... looking into why we can't use defer ...

yaacov:
@jhernand can you think of a reason we can't use defer inside the test loop ?

pkg/config/builder_test.go
yaacov: Foe some reason defer doesn't work here :-(

yaacov: defer is for functions, it doesn't work in code blocks.

yaacov:
@jhernand :fireworks: :cake: ready to squash ?

yaacov:
/test pull-autoheal-check

[ squashed the test changes to one commit ]

jhernand:
@yaacov `defer ...` also works for loops, only that all the actions that you defer in the body of the loop will be executed when you exit the function that contains the loop. If what you want is to execute the deferred actions after each iteration of the loop, then you can wrap the body of the loop inside an anonymous function:

```go
for _, test := range configsTest {
  func () {
    ...
    cfg, err := l.Build()
    if err != nil {
    }
    defer cfg.ShutDown();
    ...
  }()
}
```

Or else you can create a regular function, and call it from the loop.

I think that any of these two alternatives is better than repeating the clean up actions in all the code paths, as it is very easy to forget one of those code paths, specially in the future.

openshift-docker:
Same as origin unless we have a great reason

On May 9, 2018, at 9:38 AM, Steve Kuznetsov <notifications@github.com>
wrote:

@smarterclayton <https://github.com/smarterclayton> @jupierce
<https://github.com/jupierce> do we have any guidance as to which version
of Go to depend on for builds? My first blush impression would be to use
the same one that Origin is, but I don't really know if we have a more
specific recommendation.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/autoheal/pull/103#issuecomment-387800248>,
or mute
the thread

yaacov:
@jhernand ping ?

jhernand:
Thanks @yaacov.

/lgtm
/approved

PR #105, opened 2018-05-13
Check prompt-on-launch before launching a job

### LINKED ISSUE #89: Verify AWX `PROMPT ON LAUNCH` option before running

zgalor:
Add a check that `PROMPT ON LAUNCH` is set for limit and extra-vars in case they aren't empty.
If limit/extra-vars are configured, but the template doesn't have prompt-on-launch set for them, log a user warning.

fix #89 

prompt_on_launch

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -109,8 +109,8 @@
     "awx",
     "awx/internal/data"
   ]
-  revision = "78fe047d29a6a1fdcf7f5f0d9487706a1784b2e2"
-  version = "0.0.3"
+  revision = "5c5af1820763b6c2767f9c92fd7a3a8fe71668ff"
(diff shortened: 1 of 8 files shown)


zgalor:
@moolitayer @jhernand @yaacov @elad661 @nimrodshn Please review

jhernand:
Thanks @zgalor.

PR #106, opened 2018-05-15
Example config: clarify what keys credentialsRef secret should have

cben:
This is already documented in README but won't hurt in example config too.


moolitayer:
Thanks @cben, I think an example OR a reference to the README would be useful too. WDYT?

Issue #85, opened 2018-04-30
Use etcd to and an API server to manage the healing rules


jhernand:
Currently the healing rules are stored in the configuration file of the auto-heal service. This is OK when there only a few rules, and when reconfiguring the auto-heal service to add/remove/modify healing rules isn't frequent. But to be really useful the auto-heal service should be very easy to reconfigure, because problems and their temporary solutions are very dynamic.

One possible solution to make the auto-heal service more dynamicv is to create an API server that manages the healing rules, storing them in an etcd database. This would allow other components to add/remove/modify healing rules via this API, without having to change the configuration of the auto-heal service directly. We could even build a GUI on top of that API, to allow human users to add/remove/modify healing rules manually.

Initially this API server doesn't need to be integrated with the rest of API servers of the cluster, but it could, in the future.

I have created #84 to explore this.

moolitayer:
Version controlled heals in private repositories is consistent with how the ops worked in openshift-tools.
Maybe we should mix this with https://github.com/openshift/autoheal/issues/86 somehow? - Creating heals from the config file or taking existing ones from SCM?

Can we get to a situation where there is only one file to update when adding a new heal?

PR #107, opened 2018-05-17
[WIP] Basic etcd usage in autoheal

zeari:
Solves: #85

Added:
* An etcd pod in `template.yml`
* vendor code
* custom etcd endpoint url in config
* A `client.GetVersion` from etcd when the server starts

@elad661 @cben @nimrodshn @jhernand Please review

Also adds etcd related messages in console:
```Shell
autoheal[etcd*]$  make run-dev
hack/run-dev.sh
I0517 15:19:19.809765   13167 config.go:201] Loading configuration file 'examples/autoheal-dev.yml'
I0517 15:19:19.814616   13167 config.go:105] Watching configuration file 'examples/autoheal-dev.yml'
I0517 15:19:19.814632   13167 healer.go:136] AWX user is 'my-user'
...
I0517 15:19:19.814762   13167 healer.go:223] Trying to connect to etcd server...
I0517 15:19:19.815359   13167 healer.go:228] Etcd server version: 3.3.5 available at [http://127.0.0.1:2379]
....
I0517 15:19:19.815379   13167 healer.go:237] Web server started

```


pkg/config/builder.go
```diff
@@ -108,6 +119,9 @@ func (b *Builder) Build() (c *Config, err error) {
 		files:         b.files,
 		loadMutex:     &sync.Mutex{},
 		listenerMutex: &sync.Mutex{},
+		etcd: &EtcdConfig{
+			endpoint: "http://autoheal-etcd.openshift-autoheal.svc.cluster.local:80",
```
elad661: Is this hardcoded on purpose?

template.yml
```diff
@@ -215,3 +215,35 @@ objects:
     - name: autoheal
       port: 443
       targetPort: 8443
+
+- apiVersion: v1
+  kind: Pod
```
elad661: This should be a `Deployment`, not a static pod.

template.yml
```diff
@@ -215,3 +215,35 @@ objects:
...
+  metadata:
+    name: autoheal-etcd
+    labels:
+      app: autoheal-etcd
+  spec:
+    containers:
+    - name: etcd
+      image: quay.io/coreos/etcd:v3.3.5
```
elad661: This should be configurable

template.yml
```diff
@@ -215,3 +215,35 @@ objects:
...
+      command:
+      - "etcd"
+      - "--listen-client-urls=http://0.0.0.0:4001"
```
elad661: Can this be HTTPS?

elad661 (commented):
Few minor comments about the template changes.

Also, you probably want to put `oauth-proxy` or `kube-rbac-proxy` in the etcd pod, and configure it so that only autoheal can access this etcd. Otherwise it's exposed to all users on the cluster and that's not great from a security perspective.

zeari:
> Can this be HTTPS?

@elad661 Lets take care of that after we figure out  some basic use cases :sweat_smile:

zeari:
> When I opened #85 I was thinking on using etcd as the storage for healing rules, but indirectly, via a Kubernetes API server. In that scenario the auto-heal service would not need to access directly the etcd server, so it wouldn't need to have a parameter containing the etcd server address, but rather the API server address. Do you think that we should use etcd directly instead of via the API server?

*I think both options are fine.*
Pros for this PR: 
* managing etcd and having it customize-able with an api enpoint parameter gives us some extra control. 
* I understand that the general direction is not to have too many integrated services using the kube API

nimrodshn:
> - managing etcd and having it customize-able with an api enpoint parameter gives us some extra control. 

> - I understand that the general direction is not to have too many integrated services using the kube API
I support @zeari's approach as well for the simple reason - one can deploy autoheal without a k8s cluster, in which case, using etcd throught k8s api server would fail. But if we simply use etcd as a cache that can work without the dependency on k8s.

PR #108, opened 2018-05-28
Add a Dockerfile for simple build from source, not through rpms

cben:
I just want to run it, using openshift-ansible, so I need an image :-)
The default image specified by openshift-ansible doesn't exist.
`make build-rpms build-images` doesn't work for me as I don't have `imagebuilder`.

This uses a multi-stage build which requires docker 17.05 or higher, not available in fedora yet, works fine on Quay (probably Docker Hub too, untested).
https://quay.io/repository/cben/autoheal
`docker pull quay.io/cben/autoheal:dockerfile`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,10 +9,11 @@ RUN apk --no-cache add ca-certificates
 RUN useradd --no-create-home autoheal
 USER autoheal
 EXPOSE 9099
-WORKDIR /root/
+
+WORKDIR /bin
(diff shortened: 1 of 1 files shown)


jhernand:
@cben is this replaced by #109?

cben:
I still think a simpler build without rpms is nice in an open source project.
But as we're not maintaining the project, having 2 build ways just increases the chance they'll break...
Plus Fedora's docker is still ancient so this won't work locally (for us).

